PR #5, opened 2016-02-08
Small bugfixes and new feature

apuigsech:
I fixed two small problems; One typo on the README file, and a problem when using Default field values where the default value were never used.

I also added support to add Parameters on Field Hooks that can be useful in some cases.


rs:
Could you please split those different changes into their own PRs?

apuigsech:
That's what I tried oIt's what I tried at first instance, but failed :(

This afternoon I'll fix it.

apuigsech:
Splited into 3 PRs.

rs:
Thx

PR #8, opened 2016-02-08
Add field hook params

apuigsech:


rs:
I'm not convinced by this design. I think field hooks should be limited to the field and we may have resource hooks for global consistency check involving more than one field. With per field model, you can't really control the order of execution for instance. It does also add complexity to the API.

apuigsech:
I agree you somehow, but I think that sometimes you will need some external information on OnInit/OnUpdate hook. One simple example could be if there are some configuration setup that can affect on field initialization, currently the only way (and probably is the good one, I am not sure) to do it is using global variables.

My first approach was to create OnInit and OnUpdate hooks for Schema object and do it globally, having all fields accesible, but I had some problems to chose WHEN execute those hooks. I can recover the code and share it with you.

However it does not solve the problem of having access to some external information.

rs:
What about passing the `net/context`? You could store whatever you want in this context and use it in your hooks.

apuigsech:
Sounds great. 

I have to learn a bit more about net/context. I will explore it.

rs:
Please review this [commit](https://github.com/rs/rest-layer/commit/597ca9256a9ba88e4c5f0517337e1fdc39d4aa57) and tell me if it fulfill your needs.

Issue #23, opened 2016-07-13
Describing Arrays of Objects?


yanfali:
Hello, I'm struggling at the moment in trying to describe Arrays of 'object'. It's not obvious looking at the primitives how to do that. I have experimented with setting a `schema.Fields` with a Validator for type `schema.Array` and then using the `Field.Schema` to describe the object, but I'm not sure this would validate correctly. Ideally, what would be interesting is if I could set a `schema.Schema` on an `Array`, so one could see the hierarchy.

Would there be interest in taking a patch for `Array` that would let me nest a `Schema` within an `Array`. `Array` itself doesn't let you map key validation to value validation in an obvious way. perhaps this is a documentation issue?

Thanks for any guidance.

rs:
With `schema.Array` you can set the `ValuesValidator` to any validator. You can set this to a `schema.Dict` to describe your sub-objects for instance. Isn't it what you need?

yanfali:
I did try that, but it wasn't clear how to map the names to the values
validator. For example, if some fields are strings and some are number
wouldn't dict allow either to be valid in a particular dict?

rs:
That's true, the `schema.Dict` validator doesn't let you specify specific validator for given keys. You may either improve it or create a more advanced dict validator for that. It would be great to have this feature in the core.

yanfali:
OK, I put up a PR for you to digest, that could be one possible way to solve this problem. It heavily leverages the existing code and uses an embedded Schema instance. I'm not sure if this is the right approach, so I would love some feedback. Also if you could let me know what other things you might like to see in there I will do my best to add them.

rs:
Can we say this issue is resolved?

yanfali:
Absolutely

PR #24, opened 2016-07-15
POC: Implement an Object validator

yanfali:
Issue #23 Proof of concept

Create an "object" Validator which leverages leverages Schema validation to validate a sub document using Schema itself.

Goal is to allow you to creation arrays of objects which have their fields validated correctly.

This needs more tests and work, but I wanted to get your feedback @rs on this general approach, or if you have thoughts on a better way to do this.

### DIFF
--- a/schema/object.go
+++ b/schema/object.go
@@ -1,0 +1,55 @@
+package schema
+
+import (
+	"errors"
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// Object validates objects which are defined by Schemas.
+type Object struct {
+	Schema *Schema
+}
+
+// Compile implements Compiler interface
+func (v *Object) Compile() error {
+	if v.Schema == nil {
+		return fmt.Errorf("No schema defined for object")
+	}
+
+	if err := compileDependencies(*v.Schema, v.Schema); err != nil {
+		return err
+	}
+	return nil
+}
+
+// ErrorMap to return lots of errors
+type ErrorMap map[string][]interface{}
+
+func (e ErrorMap) Error() string {
+	errs := make([]string, 0, len(e))
+	for key := range e {
+		errs = append(errs, key)
+	}
+	sort.Strings(errs)
+	for i, key := range errs {
+		errs[i] = fmt.Sprintf("%s is %s", key, e[key])
+	}
+	return strings.Join(errs, ", ")
+}
+
+// Validate implements FieldValidator interface
+func (v Object) Validate(value interface{}) (interface{}, error) {
+	dict, ok := value.(map[string]interface{})
+	if !ok {
+		return nil, errors.New("not a dict")
+	}
+	dest, errs := v.Schema.Validate(nil, dict)
+	if len(errs) > 0 {
+		var errMap ErrorMap
+		errMap = errs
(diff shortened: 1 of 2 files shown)


yanfali:
Travis failed, but it looks like it's errors in the JWT implementation and not related to this PR.

yanfali:
@rs feedback on the direction of this PR would be appreciated. I'm holding off on JSON Schema stuff because I could use something like this to implement "items" of arrays. Thanks

rs:
I’m on vacation with limited access to internet. I will look at it asap.

yanfali:
Thanks @rs I assumed you were on vacation so was being patient.

rs:
JWT usage is fixed in master. Please rebase.

yanfali:
Weird, now it's triggering --- FAIL: TestRoutePathParentsExists (0.00s) errors :(

yanfali:
rebased. Thanks

rs:
By reading the code, I'm not sure to see how the errors would be formatted with sub-fields. Don't you think it could be a good idea to add some tests on the format of the error in different context?

yanfali:
Good point. I will do that. Thanks for the feedback.

yanfali:
OK, I've added 3 more tests. Because the error messages are a map, go makes the ordering that comes back non-deterministic, so I'm using contains to test for multiple values detected as bad. I made an array test but am only testing 1 bad value in the array because Array will stop executing after the first value that does not validate.

I also improved the error message a little to make it more readable. 1 thing I noticed which I may send a PR in for, is that `Bool` has inconsistent casing for it's error message. Everything else reports lower case except `bool.go`

Thanks.

rs:
Boolean does always take a capital as it is coming from a proper noun.

http://english.stackexchange.com/questions/4481/should-the-word-boolean-be-capitalized

schema/object.go
```diff
@@ -0,0 +1,53 @@
...
+	return strings.Join(errorStrings, ", ")
+}
+
+// Validate implements FieldValidator interface
+func (v Object) Validate(value interface{}) (interface{}, error) {
+	dict, ok := value.(map[string]interface{})
+	if !ok {
+		return nil, errors.New("not a dict")
+	}
+	fmt.Printf("%+v", value)
```
rs: Is it a debug leftover?

rs:
What about something like that to ease testing: https://play.golang.org/p/Dn5htsZKSv

schema/object.go
yanfali: yep thinko. I'll remove it.

yanfali:
This looks good. I was avoiding making the Error() function more complex, but if you're ok with a sort and the extra garbage I'll use your implementation. Thanks!

yanfali:
OK, rebased, removed thinko and refactored ErrorMap and tests to be like much nicer playground example. Thanks for all the feedback!

yanfali:
OK, I spotted an unnecessary allocation and got rid of it in the error path.

yanfali:
Fixed an ordering issue in arguments to assert.Equal

rs:
Thx @yanfali you're awesome

PR #25, opened 2016-07-18
Feature: specify time formats we wish to validate

yanfali:
- specify an exact list of time formats we wish to validate via
  TimeLayouts member.
- update unit tests to Compile before using instance

### DIFF
--- a/schema/time.go
+++ b/schema/time.go
@@ -52,12 +52,28 @@ var (
 
 // Time validates time based values
 type Time struct {
+	TimeLayouts []string // TimeLayouts is set of time layouts we want to validate
+	layouts     []string
+}
+
+// Compile the time formats
+func (v *Time) Compile() (err error) {
+	if len(v.TimeLayouts) == 0 {
+		// default layouts to all formats
+		v.layouts = formats
+		return nil
+	}
+	// User specified list of time layouts
+	for _, layout := range v.TimeLayouts {
+		v.layouts = append(v.layouts, string(layout))
+	}
+	return nil
 }
 
 // Validate validates and normalize time based value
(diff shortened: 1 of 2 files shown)


yanfali:
@rs looks like xlog is no longer building on travis, which appears to be causing failures in rest-layer. :(

yanfali:
Going to close this temporarily while I debug why it's not implementing the `schema.Compiler` interface.

yanfali:
OK, re-opening. It was my use of this interface. Validators need to be pointers to instance for the interface cast to work.

rs:
I fixed xlog but the tests are still failing on jwt example.

rs:
I fixed JWT usage (JWT API changed).

schema/time.go
```diff
@@ -50,14 +51,46 @@ var (
...
+
+// validTimeLayout Validates user specified time layouts against
+// formats.
+func validTimeLayout(tf TimeLayout) error {
+	for _, layout := range formats {
+		if layout == string(tf) {
+			return nil
+		}
+	}
+	return fmt.Errorf("unknown time layout %q", tf)
```
rs: Why do you restrict layouts to pre-defined ones?

yanfali: Was trying to keep it restricted to same set as it normally expected. Am
happy to remove restriction for slightly less safety but more flexibility.

rs: I'm all for flexibility :)

yanfali: I'll fix it and resubmit tonight. Thanks

yanfali:
Rebased and removed excess validation.

PR #32, opened 2016-08-19
Thinko. Actually put the files in the hierarchy

yanfali:
- a big mistake on my part was not testing the code outside of the
   path. This commit fixes the hierarchy and adds a test to ensure that
   the import path functions.


schema/encoding/encoder_test.go
```diff
@@ -0,0 +1,24 @@
+package encoding
```
rs: Shouldn't tests be in the same package as jsonschema?

yanfali:
In this case no, because it's specifically to test that importing paths work correctly. The mistake I made earlier was to not try and import from another package, but only use it from the same package :(

rs:
Got it. That's why I always happened `_test` to my test package :)

yanfali:
Thanks!

rs:
Forget what I said, I must have dreamed it :)

PR #38, opened 2016-09-09
Fix Object as a Field use case

yanfali:
I had not tested the object as a field use case thoroughly enough and
encountered this when I used it.

If you use Object as a Field of a Schema rather than as embedded in an
Array, it's original use case, the returned sub schema is wrapped in an
object. This creates invalid JSON.

 - modify the internal functions validatorToJSONSchema and
   schemaToJSONSchema to accept an object flag. If the flag is true then
   return the JSONSchema wrapped in an object. If false return it
   unwrapped.
 - the object flag to false when invoked from validatorToJSONSchema
 - the object flag is true when invoked from schema.Array handler
 - this also guards against nil in the schema.Object Schema field
 - cover array case with nil schema

### DIFF
--- a/schema/encoding/jsonschema/jsonschema.go
+++ b/schema/encoding/jsonschema/jsonschema.go
@@ -42,7 +42,7 @@ func (ew errWriter) write(b []byte) {
 }
 
 // ValidatorToJSONSchema takes a validator and renders to JSON
-func validatorToJSONSchema(w io.Writer, v schema.FieldValidator) (err error) {
+func validatorToJSONSchema(w io.Writer, v schema.FieldValidator, object bool) (err error) {
 	if v == nil {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


schema/encoding/jsonschema/jsonschema.go
```diff
@@ -101,12 +101,20 @@ func validatorToJSONSchema(w io.Writer, v schema.FieldValidator) (err error) {
 		if t.ValuesValidator != nil {
 			ew.writeString(`, "items": `)
 			if ew.err == nil {
-				ew.err = validatorToJSONSchema(w, t.ValuesValidator)
+				ew.err = validatorToJSONSchema(w, t.ValuesValidator, true)
```
smyrman: why not.
ew.writeString("{")
ew.err = validatorToJSONSchema(w, t.ValuesValidator)
ew.writeString("}")

Then you don't need the extra object flag parameter...

yanfali: don't know if it covers all the use cases, if you can work a patch that makes it work though that'd be great.

smyrman: While looking into that, I think I found a bug -- seams you put the "required" list inside "properties". This is incorrect, it should be outside (same level as "type": "object"). Otherwise it would mask out any field named "required".

See Basic Example:
- http://json-schema.org/examples.html

Sorry I missed this in the initial review!

yanfali: Oh good catch. I'll file an issue and see if can't fix it.

Issue #40, opened 2016-09-09
JSONSchema required


yanfali:
When schema is embedded within an Array or an Object Field the required field is being added to the "items" or "properties" body. This should appear one level up at the same level as "type"
Reported by @smyrman

Example of incorrect output

```
        "items": {
          "properties": {
            "key": {
              "type": "string"
            },
            "score": {
              "type": "integer"
            },
            "required": [
              "score",
              "key",
              "value"
            ],
            "value": {
              "type": "string"
            }
          },
          "type": "object"
        },
```

Example of correct output

```
        "items": {
          "properties": {
            "key": {
              "type": "string"
            },
            "score": {
              "type": "integer"
            },

            "value": {
              "type": "string"
            }
          },
          "required": [
            "score",
            "key",
            "value"
          ],
          "type": "object"
        },
```

yanfali:
@rs I'll try and work up a patch, I have an idea of how to solve this.

yanfali:
@smyrman you might want to take a look now. I think it does the right thing. It's definitely generating valid JSON.

smyrman:
From what little testing I could do, it looks fine, at least inside an items clause.

smyrman:
@rs, @yanfali, I assume this issue should be closed.

PR #41, opened 2016-09-10
Fix JSONSchema for Embedded Objects

yanfali:
When embedding objects within a JSON Schema their Required field should
not be embedded within the "properties" field, but one level above with
the type information.

 - add unit tests to test object and array with embedded object schema
 - do not generate required fields within embedded object schemas
 - generate required fields for array items and object fields

This should fix issue #40


yanfali:
missed a corner case

PR #43, opened 2016-09-10
Fix JSONSchema for Embedded Objects

yanfali:
When embedding objects within a JSON Schema their Required field should
not be embedded within the "properties" field, but one level above with
the type information.

 - add unit tests to test object and array with embedded object schema
 - required fields generation has always been in wrong location
   - move it outside of properties
 - move object notation outside of schemaToJSONSchema into higher level
   public method to make it re-usable


yanfali:
OK, third time's the charm. This one takes @smyrman's advice and removes the extra flag and also fixes the required array. It's always been in the wrong place, embedded in properties.

yanfali:
beefed up the unit test to check for required field at the correct level in schema

Issue #44, opened 2016-09-13
[jsonschema] Table driven tests


smyrman:
Hi.

I am interested in contributing some table driven tests for the `jsonschema` package. I am picturing an array of schema definitions that can be compared to a JSON string via [asserts.JSONEq](https://godoc.org/github.com/stretchr/testify/assert#JSONEq). I would expect my progress to be a bit slow, as it's a busy period for me familywise.

@rs, is depending on Go 1.7 sub-tests (inside the test package only) OK, or should I write my tests so they work for Go 1.6?

@yanfali, I though I would mention your name as well so you get notified ;-)

rs:
Go for 1.7 only, just add a build flags so it won't break 1.6 CI.

smyrman:
Started of with some simple cleanups first in #45. @yanfali, are you also happy with these?

smyrman:
> Go for 1.7 only, just add a build flags so it won't break 1.6 CI.

Actually, when I started coding, I found it was easy enough to support both (via build tags) -- I will use parallel sub-tests for Go 1.7 and a sequential approach without sub-tests for Go < 1.7.

My initial work lives on [this branch](https://github.com/smyrman/rest-layer/tree/table-tests).

It's based on top of #45. I will push it to a PR soonish.

smyrman:
Initial review for table-tests added in #46. @rs, when you have time, let me know if this is moving in the right direction, and what you would like to change.

yanfali:
I thought this was merged. I used the pattern recently in one of my projects.

smyrman:
I didn't convert all the jsonschema test, but yeah, I think we can close it.

rs:
Kind of unrelated, but shouldn’t we move all the jsonschema tests inside the package?

smyrman:
Sounds sensible

smyrman:
>  shouldn’t we move all the jsonschema tests inside the package?

@rs, I am now doing this in #64. This commit also moves the remaining tests into table tests when appropriate.

PR #45, opened 2016-09-14
JSON Schema: Cleanup and documentation

smyrman:
Cleaned up the JSON Schema package documentation a bit, as well as some
of the code. I changed one of the tests and the ErrNotImplemented error
message. Pacakge impors was rearanged by goimports.

I used the Go standard library as a guide line for the docs, including
the use of full sentences with a trailing period in comments and short,
lower case error messages.

### DIFF
--- a/schema/encoding/jsonschema/jsonschema.go
+++ b/schema/encoding/jsonschema/jsonschema.go
@@ -4,15 +4,17 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/rs/rest-layer/schema"
 	"io"
 	"strconv"
 	"strings"
+
+	"github.com/rs/rest-layer/schema"
 )
 
 var (
-	// ErrNotImplemented means schema.FieldValidator type is not implemented
-	ErrNotImplemented = errors.New("Schema not Implemented")
+	// ErrNotImplemented is returned when the JSON schema encoding logic for a schema.FieldValidator has not (yet)
+	// been implemented.
+	ErrNotImplemented = errors.New("not implemented")
 )
 
 type errWriter struct {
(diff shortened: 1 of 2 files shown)


schema/encoding/jsonschema/jsonschema.go
```diff
@@ -166,29 +168,32 @@ func schemaToJSONSchema(w io.Writer, s *schema.Schema) (err error) {
...
-	ew.writeFormat(`, "required": [%s]`, strings.Join(required, ", "))
+
+	if len(required) > 0 {
+		ew.writeFormat(`, "required": [%s]`, strings.Join(required, ", "))
+	}
 	return ew.err
 }
 
-// Encoder encodes schema.Schema into a JSONSchema
+// Encoder writes the JSON Schema representation of a schema.Schema to an output stream. Note that only a sub-set of the
```
smyrman: Broke the lines at 120 chars. @rs, is there any particular line width you prefer or use elsewhere in this project?

yanfali:
That looks pretty good. Is the required array optional if empty?

schema/encoding/jsonschema/jsonschema_test.go
```diff
@@ -413,42 +414,38 @@ func TestArrayOfObjects(t *testing.T) {
 			},
 		},
 	}
+	const expect = `{
```
yanfali: much cleaner

smyrman:
> That looks pretty good. Is the required array optional if empty?
I thought so, but let me double check it...

smyrman:
>> That looks pretty good. Is the required array optional if empty?

> I thought so, but let me double check it...

Yes, looks like It is optional.

According to the [Core/Validation Meta-Schema](http://json-schema.org/schema) (which is the JSON Schema for JSON Schema), there are no required fields for a "JSON Schema" -- even "type" is optional.

yanfali:
LGTM. Do you want to add a test for required not being in output?

smyrman:
Sure, let's add it here.

Other than that, I hope to add more tests through #44 :-)

smyrman:
> Sure, let's add it here.

Actually, we already have it! Notice that the outer schema  in `TestArrayOfObjects()` does not have the required array...
```
const expect = `{
	"type": "object",
	"title": "A list of students",
	"properties": {
		"students": {...}
	}
	// <-- no required list
}`
```

I can add it as a separate test in #44.

yanfali:
Sounds good

smyrman:
@rs, this one is ready if you are happy with it.

schema/encoding/jsonschema/jsonschema.go
rs: I don't particularly care, 120 is fine.

PR #46, opened 2016-09-16
Table tests (initial code)

smyrman:
Pushing for an initial review to check that I am moving in the right direction. Note that this PR is based on #45, so we should probably either merge that first, or abandon it and review everything here. @rs, up to you.

I was aiming to move in a similar direction as the schema package by having one test file per Validator type. Even there is no file per validator for the encoding part yet (maybe there will be in the future to prevent validatorToJSONSchema from growing to large?), it made it easier to see what I have implemented tests for, and to splitt concerns...


smyrman:
Tested Go 1.6 code-path and pushed an amended commit to the branch:

```
$ docker run -v $GOPATH:/go golang:1.6 go test -v github.com/rs/rest-layer/schema/encoding/jsonschema
...
ok  	github.com/rs/rest-layer/schema/encoding/jsonschema	0.019s
```

schema/float_test.go
```diff
@@ -28,6 +29,18 @@ func TestFloatValidator(t *testing.T) {
...
+	assert.Equal(t, 3.1, s)
+	s, err = Float{Boundaries: &Boundaries{Min: math.NaN(), Max: 10}}.Validate(3.1)
+	assert.NoError(t, err)
+	assert.Equal(t, 3.1, s)
+	s, err = Float{Boundaries: &Boundaries{Min: 2, Max: math.Inf(1)}}.Validate(3.1)
+	assert.NoError(t, err)
+	assert.Equal(t, 3.1, s)
+	s, err = Float{Boundaries: &Boundaries{Min: 2, Max: math.NaN()}}.Validate(3.1)
+	assert.NoError(t, err)
+	assert.Equal(t, 3.1, s)
```
smyrman: Not sure if support for NaN and Inf values was intended to work, but since it does work, I added some tests for it.

schema/integer_test.go
```diff
@@ -28,6 +29,18 @@ func TestIntegerValidator(t *testing.T) {
...
+	assert.Equal(t, 3, s)
+	s, err = Integer{Boundaries: &Boundaries{Min: math.NaN(), Max: 10}}.Validate(3)
+	assert.NoError(t, err)
+	assert.Equal(t, 3, s)
+	s, err = Integer{Boundaries: &Boundaries{Min: 2, Max: math.Inf(1)}}.Validate(3)
+	assert.NoError(t, err)
+	assert.Equal(t, 3, s)
+	s, err = Integer{Boundaries: &Boundaries{Min: 2, Max: math.NaN()}}.Validate(3)
+	assert.NoError(t, err)
+	assert.Equal(t, 3, s)
```
smyrman: Same here. Not sure if support for NaN and Inf values was intended to work, but since it does work, I added some tests for it.

schema/encoding/jsonschema/jsonschema.go
```diff
@@ -41,7 +44,25 @@ func (ew errWriter) write(b []byte) {
...
+	ew := errWriter{w: w}
+
+	if !math.IsNaN(b.Min) && !math.IsInf(b.Min, -1) {
+		ew.writeFormat(`, "minimum": %s`, strconv.FormatFloat(b.Min, 'E', -1, 64))
+	}
+	if !math.IsNaN(b.Max) && !math.IsInf(b.Max, 1) {
+		ew.writeFormat(`, "maximum": %s`, strconv.FormatFloat(b.Max, 'E', -1, 64))
+	}
+	return ew.err
+}
```
smyrman: I added this method to use the same code-path for encoding both integer and float boundaries. I also added support for NaN and Inf values so that:

- Max=NaN is treated as no limit.
- Max=+Inf is treated as no limit.
- Min=NaN is treated as no limit.
- Min=-Inf is treated as no limit.

Might be useful in some cases to only set one limit... I checked that this worked with the existing schema package logic by first inspecting the code, and then by adding some tests.

Hint: The following is always true for NaN, where f is any float64 value:
- `(f < NaN) == false`
- `(f > NaN) == false`

rs:
Can you please rebase on master?

smyrman:
rebased

schema/encoding/jsonschema/all_test.go
```diff
@@ -0,0 +1,216 @@
...
+							},
+						},
+					},
+				},
+			},
+			// FIXME: "object" is the only field-type that encodes description as "title" instead of
+			// "description". Should it be encoded as "description" here as well?
+			expect: `{
+				"type": "object",
+				"title": "A list of students",
```
smyrman: I am wondering if we should switch this to "description". @yanfali, do you have an opinion? Is it any reason in particular that you chose to encode the schema description as "title" for "object", while it's encoded as "description" for other types?

JSON Schema defines both "title" and "description", where title is supposed to be shorter, and description is supposed to be longer.

smyrman:
Added float tests, added one more "all" (package level) test, and removed redundant tests from old test file.

schema/encoding/jsonschema/jsonschema_test.go
```diff
@@ -261,232 +205,8 @@ func TestObjectWithNilSchema(t *testing.T) {
...
-		}
-	}`
-
-	b := new(bytes.Buffer)
-	encoder := NewEncoder(b)
-	assert.NoError(t, encoder.Encode(s))
-	assert.JSONEq(t, expect, b.String())
-}
-
-func TestDefaultEncodingWithStringFieldAndIntegerDefault(t *testing.T) {
```
smyrman: OK, haven't added any tests for default, so this was removed by mistake. Will add some tests to replace it.

schema/encoding/jsonschema/all_test.go
smyrman: I think the best thing to do would be to switch to "description", and if we need "title" in the future, add a Title property to schema.Field.

yanfali:
Sorry I've been tied up at work. I will review it this evening PST and let you know my thoughts. Thanks for doing this.

smyrman:
Amended to include tests for Default.

smyrman:
> Sorry I've been tied up at work.

No worries:-)

schema/encoding/jsonschema/jsonschema_test.go
smyrman: done

schema/encoding/jsonschema/all_test.go
yanfali: I don't feel strongly about this. In fact the first time I submitted this, I totally forgot I used title.

schema/encoding/jsonschema/all_go16_test.go
```diff
@@ -0,0 +1,11 @@
+// +build !go1.7
+
+package jsonschema_test
```
yanfali: interesting, why does this have a separate package? I'm not familiar with this pattern. Doesn't this mean you can't inspect package variables? Is that the intention? what happens when you run this from the top most level via ./... ?

schema/encoding/jsonschema/all_test.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+// Run runs the testCase according to your Go version. For Go >= 1.7, test cases are run in parallel using the Go 1.7
+// sub-test feature. For older versions of Go, the testCase name is simply logged before tests are run sequentially.
+func (tc *encoderTestCase) Run(t *testing.T) {
+	tc.run(t)
+}
+
+// test performs the actual test and is used by both implementations of run.
+func (tc *encoderTestCase) test(t *testing.T) {
```
yanfali: neat

schema/encoding/jsonschema/all_test.go
```diff
@@ -0,0 +1,324 @@
...
+					}
+				}
+			}`,
+		},
+		{
+			name: `type(Default)=map[string]interface{}`,
+			schema: schema.Schema{
+				Fields: schema.Fields{
+					"name": schema.Field{
+						Default: map[string]interface{}{
```
yanfali: ok, this is an odd use case, is this the same as allowed?

schema/encoding/jsonschema/jsonschema.go
yanfali: hmm, isn't NaN an error?

smyrman: NaN means Not a Number

schema/encoding/jsonschema/all_test.go
smyrman: In which case I suggest we change it to description for consistency :-)

schema/encoding/jsonschema/jsonschema.go
yanfali: Yes but why accept it? I mean, nan is triggered by strings, so is it a good idea to silently accept it?

schema/encoding/jsonschema/all_go16_test.go
smyrman: @yanfali, This is the Go way to test the public interface of your package. It makes you having to import the package you are testing, and only give access to the public interface. Presumably you could have done the same for the tests you put in the encoding package.

schema/encoding/jsonschema/all_test.go
smyrman: Yeah, just documenting the current behaviour. I adde a comment some way up (above the test where I use an int).

schema/encoding/jsonschema/all_go16_test.go
yanfali: Excellent, thanks for explaining. The encoding test was testing imports rather than public interface so the intention was different. I was ignorant of this pattern.

smyrman: I was hoping to link to a blog post or some docs explaining the "_test" suffix, but I couldn't find it just now. However, it's a very common pattern, and you can also see it's commonly used in the Go standard library. E.g. https://golang.org/src/strings/reader_test.go

smyrman:
> hmm, isn't NaN an error?

In many programs it probably is, but for Go it's a IEEE 754 “not-a-number” value. I found it "fitting" as a way of descriing "no limit", although it's not strictly needed in our case as we could rely on +/-Inf as the only "supported" values. However, without special-casing, both NaN and Inf numbers would result in invalid JSON if specified by the user.

smyrman:
> Only major concern is package name. I made a similar choice when I created the jsonschema package, so make sure it runs from the top most level too. Putting it in subpath test/ would probably work just as well.

As mentioned, adding a "_test" suffix to the package name for any test that do not require access to private methods or variables, is the Go way. Also, it's the Go way to keep all package tests in the same folder/package so that e.g. "go test -cover <import path>" never lies;-)

schema/encoding/jsonschema/all_test.go
smyrman: I will change it to be less odd to avoid confusion, probably by setting the default one level out.

smyrman: done

schema/encoding/jsonschema/all_test.go
smyrman: done

smyrman:
Latest change include some encoding changes:
- Renamed "title" ->  "description" for consistency on object.
- Set "additionalProperties".

Test changes:
- Moved string tests from old test file to string_test.go
- Reworked float/integer tests to compare only a sub-set of the JSON (The same way string_test.go now does it...).

schema/encoding/jsonschema/jsonschema.go
smyrman: > I mean, nan is triggered by strings

What do you mean it's triggered by strings? Where, and how? E.g `strconv.ParseFloat()` would return an error and a zero value, not NaN: https://play.golang.org/p/Xx50p-LgL7

> so is it a good idea to silently accept it?

I don't think there are too many ways people can set NaN by mistake. They may get an `NaN` if they do `math.Acos(1.1)` etc., but that doesn't really sound like a common pattern for setting float limits? I would assume the most reasonable way people could set a NaN value, would be if they did so deliberately via `math.NaN()`...

> Yes but why accept it?
Currently I made the JSON Schema accept it simply because the `schema` package already accepts it (silently), and we should handle the value somehow....

smyrman: PS! This doesn't mean it can't be changed, if there is a compelling reason to change how NaN is handled by the `schema` package.

rs:
@smyrman: Are you ready for merging?

smyrman:
@rs, yes, I think this PR can be merged.

rs:
Could you please rebase on master?

smyrman:
@rs, it's now rebased again.

PR #47, opened 2016-09-21
Add additional properties key

yanfali:
Schema has a strict interpretation of the schema, and is explicit about
 what it accepts. A conforming JSONSchema document should set
 "additionalProperties" to false to document this behavior.

 This documents the existing behavior of schema, which is do not accept
 parameters which have not been specified.

 @see http://json-schema.org/latest/json-schema-validation.html#anchor64


yanfali:
@smyrman one of my co-workers pointed out it would probably be best to set this value, as it's basically the behavior of schema.

smyrman:
I thought it defaultet to false, but I see now that it does not http://json-schema.org/latest/json-schema-validation.html#anchor64

Thanks

smyrman:
Looks good to me!

yanfali:
@rs can you take a look? thanks

yanfali:
thank you!

Issue #50, opened 2016-11-29
if there is no Validator associated with a schema Field it generates illegal json


yanfali:
e.g. 			"authentication": schema.Field{
				Description: "Ignored authentication token",
				Required:    false,
			},

Is perfectly valid go code, and the validator runs without complaint, but for the JSON schema it generates illegal JSON Schema. Should we make this a run time error if we discover it in the JSONSchema or in schema?

rs:
Why is it invalid?

yanfali:
Extra comma. I think we can fix that part but I'm wondering if we shouldn't
flag it as an error? Is a schema field without a validator?

rs:
Wouldn't it be useful to let a field be of any type/format?

yanfali:
Not sure that's legal in json schema and also defeats the point of schema
as a validator. I'm not sure what the use case you have in mind is.

rs:
No particular use-case. We should certainly throw an error, you're right.

smyrman:
It's perfectly legal to have a field of an arbitrary type in JSON Schema - even an empty dictionary is a valid JSON Schema!

smyrman:
I would recommemd dealing with the comma situation, as the way we add commas feels quite hard-coded and as this issue proves, perhaps a bit error prone. Would it be possible to have anorther look at what the JSON package is doing? Or perhaps even utilize the JSON package for  tag serialization?

One usecase I see for not specifying a validation for one particular field in a Schema, would be a logging API, or perhaps a job queue with arbitrary input parameters.

smyrman:
Thinking out loud, perhaps one way could be to make it the responsibility of the "writer" (e.g. call it objectEncoder, arrayEncoder) to know when an object start, printe the leading/trailing { or [  and the commas between key/value pairs.

yanfali:
I think fixing the bug first would be good, then if you feel strongly enough to rewrite the serializer then open a separate issue. While it may be legal to generate a typeless schema I'm not really sure what that purpose it would serve other than to add to confusion.

yanfali:
My initial point is that using schema without a validator appears to be an incorrect usage of the API and should probably be considered an error by something like the JSONSchema. If you want the validator to be a wildcard wouldn't something like AnyOf be a better choice?

@rs do you consider a missing validator as an incorrect API usage or a valid use case?

rs:
[...]

PR #55, opened 2016-12-10
Issue #50 detect invalid fields

yanfali:
1st commit updates schema to detect condition where both the Validator and the Schema haven't been set and return an error.

2nd commit fixes up the Encoder so that it is more robust when emitting field properties.


yanfali:
@smyrman @rs it's not the prettiest implementation, so feedback as always appreciated. Passing all tests though.

yanfali:
OK need to review failing tests.

Issue #51, opened 2016-12-01
[jsonschema]  Rethink how jsonschema deals with encoding internally


smyrman:
This issue is strongly linked to #35.

There are a couple of issues with the current implementation of the `jsonschema` package:
1. Encoding for different `Validator` types is all done by one big switch-case in one function. This has a extensability and readability problem.
2. The encoder function is "hard-coded" to add commas in the correct places, which has proven error prone, e.g. when using nil validators, as showed in #50. Sidenote: For the specific example, it still need to be discussed weather nil validators for a field is and/or should be accepted by the restlayer API or not, but as I pointed out in #50, there are real use-cases where it would be useful to accept it.
3. The interface for end-users to implement for their _own_ custom `Validator` implementations proposed in #35, currently proposing to "mirror" the current internal behavior, require the end-user to serialize a partial JSON object (i.e. exluding the {, }), and correctly deal with commas and white-space.

I propose having a look at how to better implement a more easy-to-use interface for #35, and then use the same interface internally. I also propose to split the validator encoder function, so that there is one encoder type and/or function per supported validaor type. Each of these encoders will get a separate file, mirroring the layout of the `schema` package.

I am happy coming up with a proposal/PR as I get time to think deeper about it and try out some code..

yanfali:
I think this is worth attempting. I just took a look at https://golang.org/src/encoding/json/encode.go and it's pretty much switch statements and writing to buffers, but if you can make the code less error prone that's a worth goal. Writing serializers and deserializers tends to be a low level and messy business.

smyrman:
> and it's pretty much switch statements and writing to buffers
Yeah, but they also have a separate type for each encoder (intEncoder, uintEncoder etc.). See  https://golang.org/src/encoding/json/encode.go#L391.

> Writing serializers and deserializers tends to be a low level and messy business.
I have actually written a binary one, and yes I agree. The biggest question is which solution to go with... I have three (new) proposals that I have added to #35, as I am thinking about using the same interface externally (public) as we use internally.

Each have some advantages and disadvantages.

smyrman:
FYI, I now have a working plan on this, which involves 3 steps.

First of all, I will create some micro-benchmarks for the `jsonschema` package, and add this as a separate PR.

[...]

Issue #35, opened 2016-08-22
[jsonschema] Support custom Validators


smyrman:
One of the strengths of the `rest-later/schema` package, is that users can define their own Validators. However, doing so would lead to their validators not being handled by the `schema/encoding/jsonschema` package at the moment.
# Solution 1: Marshaler interface

One solution would be to add an interface in the `schema/encoding/jsonschema` package similar to what one would find in the `encoding/json` package from the standard libs.

``` go
type Marshaler interface {
        MarshalJSONSchema() ([]byte, error)
}
```

One could also chose to be a bit different, and go for an interface that allows directly passing in the writer to use:

``` go
type Marshaler interface {
        MarshalJSONSchema(w io.writer) error
}
```

Either way, the `jsonschema` package should test if this interface is implemented for any given validator, and prefer to use it if it is.

The second option _might_ be simplest in terms of allowing users to extend existing validators. E.g. someone might write something like:

``` go
type SizedDict struct{
        schema.Dict
        MaxValues, MinValues int
}

func (sd SizedDict) Validate(value interface{}) (interface{}, error) {
        value, err := sd.d.Validate(value)
         if err != nil {
                  return nil, err
         }
         ...
}

func (sd SizedDict) MarshalJSONSchema(w io.Writer) error {
        enc = jsonschema.NewEncoder(w)
        if err := enc.Encode(sd.d); err != nil {
                return err
        }
        ...
}
```

If the first option is chosen for the interface, it might be beneficial to define an equivalent to [json.Marshal](https://golang.org/pkg/encoding/json/#Marshal) in the jsonschema package to simplify user extensions.

smyrman:
As mentioned in #51, I suggest we rethink how the end-user interface could be made easier to use.

smyrman:
The biggest disadvantage with the proposed solution 1, is that the end-user must correctly handle commas, white-space, quotes and other JSON-Encoding logic. This can be tedious and error prone. 
After some consideration, I have thought through three different solutions, where I have landed on being in favour of the most simple, stupid one, branded "Solution 2".

As mentioned in #51 this will affect how we shold do encoding internaly as well.

# Solution 2: Return map[string]interface{} (recommended)

This proposed solution involves defining a simple interface for custom validator implementations to implement:

```go
// Encoder must be implemented by custom schema.Validator implementations in order to allow
// JSONSchema serialization.
type Encoder interface {
        // JSONSchemaDraft4 should return a map containing valid JSON Schema keys and values
[...]

PR #56, opened 2016-12-10
Issue #50 fix up JSONSchema to not emit invalid JSON

yanfali:
Fixes up the Encoder so that it is more robust when emitting field properties.

I tried patching schema to detect nil Validator and Schema but this broke a lot of unit tests so I'm withdrawing that part of the patch.

@smyrman @rs this probably needs some review, but it does pass the tests inside of schema. Having problems running the tests involving MemoryHandler

```
rest/method_item_delete_test.go:23: cannot use s (type *mem.MemoryHandler) as type resource.Storer in argument to index.Bind:
        *mem.MemoryHandler does not implement resource.Storer (wrong type for Clear method)
                have Clear("golang.org/x/net/context".Context, *resource.Lookup) (int, error)
                want Clear("context".Context, *resource.Lookup) (int, error)
```

I probably have an old version in my GOPATH.

### DIFF
--- a/schema/encoding/jsonschema/jsonschema.go
+++ b/schema/encoding/jsonschema/jsonschema.go
@@ -27,20 +27,21 @@ type errWriter struct {
 	properties int       // track properties written
 }
 
-// Comma optionally outputs a comma.
+// comma optionally outputs a comma.
 // Invoke this when you're about to write a property.
 // Tracks how many have been written and emits if not the first.
-func (ew *errWriter) Comma() {
+func (ew *errWriter) comma() {
 	if ew.properties > 0 {
 		ew.writeString(",")
 	}
 	ew.properties++
 }
 
-func (ew *errWriter) ResetPropertiesCount() {
+func (ew *errWriter) resetPropertiesCount() {
 	ew.properties = 0
 }
 
+// Compatibility with io.Writer interface
 func (ew errWriter) Write(p []byte) (int, error) {
 	if ew.err != nil {
 		return 0, ew.err
@@ -62,7 +63,7 @@ func (ew errWriter) writeString(s string) {
 	_, ew.err = ew.w.Write([]byte(s))
 }
 
-func (ew errWriter) write(b []byte) {
+func (ew errWriter) writeBytes(b []byte) {
 	if ew.err != nil {
 		return
 	}
@@ -163,15 +164,15 @@ func validatorToJSONSchema(w io.Writer, v schema.FieldValidator) (err error) {
 func serializeField(ew errWriter, key string, field schema.Field) error {
 	ew.writeFormat("%q: {", key)
 	if field.Description != "" {
-		ew.Comma()
+		ew.comma()
 		ew.writeFormat(`"description": %q`, field.Description)
 	}
 	if field.ReadOnly {
-		ew.Comma()
+		ew.comma()
 		ew.writeFormat(`"readOnly": %t`, field.ReadOnly)
 	}
 	if field.Validator != nil {
-		ew.Comma()
+		ew.comma()
 		ew.err = validatorToJSONSchema(ew, field.Validator)
 	}
 	if field.Default != nil {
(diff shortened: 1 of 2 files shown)


schema/encoding/jsonschema/jsonschema.go
```diff
@@ -18,9 +18,34 @@ var (
...
-	err error
+	w          io.Writer // writer instance
+	err        error     // track errors
+	properties int       // track properties written
+}
+
+// Comma optionally outputs a comma.
+// Invoke this when you're about to write a property.
+// Tracks how many have been written and emits if not the first.
+func (ew *errWriter) Comma() {
```
smyrman: Although it does not matter to much as the errWriter type is private, shoul this method still be private for consistency with writeFormat?

For a public interface, I would use a bit more time on defining a better interface, as can be discussed further in #35 and #51.

schema/encoding/jsonschema/jsonschema.go
```diff
@@ -18,9 +18,34 @@ var (
...
+	if ew.properties > 0 {
+		ew.writeString(",")
+	}
+	ew.properties++
+}
+
+func (ew *errWriter) ResetPropertiesCount() {
```
smyrman: should this one be private as well?

schema/encoding/jsonschema/jsonschema.go
```diff
@@ -135,6 +160,34 @@ func validatorToJSONSchema(w io.Writer, v schema.FieldValidator) (err error) {
...
+		ew.err = validatorToJSONSchema(ew, field.Validator)
+	}
+	if field.Default != nil {
+		b, err := json.Marshal(field.Default)
+		if err != nil {
+			return err
+		}
+		ew.Comma()
+		ew.writeString(`"default": `)
+		ew.write(b)
```
smyrman: here you call `ew.write()`, but above you define `func 
(ew errWriter) Write(...)`?

schema/encoding/jsonschema/jsonschema.go
```diff
@@ -155,29 +208,11 @@ func schemaToJSONSchema(w io.Writer, s *schema.Schema) (err error) {
...
-		if field.Default != nil {
-			b, err := json.Marshal(field.Default)
-			if err != nil {
-				return err
-			}
-			ew.writeString(`, "default": `)
-			ew.write(b)
-		}
-		ew.writeString("}")
-		if ew.err != nil {
```
smyrman: I very much like that you move this into a new function `serializeField()`.

schema/encoding/jsonschema/jsonschema.go
```diff
@@ -18,9 +18,35 @@ var (
...
+func (ew *errWriter) comma() {
+	if ew.properties > 0 {
+		ew.writeString(",")
+	}
+	ew.properties++
+}
+
+func (ew *errWriter) resetPropertiesCount() {
+	ew.properties = 0
+}
```
rs: I'm not sure about those two methods on a type called `errWriter`. Also, I don't see the relation between `comma` and `resetPropertiesCount` semantically, those two method names do not show that they are working together.

As you only use this logic in `serializeField`, why don't you just move it inside this method? Actually you are even forced to do that, as I think the current implementation introduce a bug when the following sequence of call happens: `schemaToJSONSchema` > `serializeField` > `validatorToJSONSchema` > `schemaToJSONSchema` > `serializeField`. As you have two `serializeField` in flight at the same time, their states (which is global to the writer) will mess up with each other.

yanfali:
You are correct. I'll rethink the approach.

smyrman:
> You are correct. I'll rethink the approach.

@yanfali, If you are going to rethink the approach, please feel welcome to join the discussion in #35 as well. I apologize for the wall of text and list of _different_ proposals.

You might want to check out "Solution 3", as it also proposes an encoder that handles commas. It has some key differences though:
- Commas are handled implicitly.
- Sub-encoders must be used (only) when encoding sub-objects or arrays.
- It's a slightly bigger change.

yanfali:
Thanks @smyrman but I have a simpler approach that should address @rs
concern. I'll leave #35 to solve the larger problem.

yanfali:
@rs this should address the concurrency and API concerns.

PR #61, opened 2016-12-13
Provide: offset query param to skip query results

omani:
* counterpart to https://github.com/rs/rest-layer-mongo/pull/4

### DIFF
--- a/resource/item.go
+++ b/resource/item.go
@@ -24,11 +24,16 @@ type Item struct {
 
 // ItemList represents a list of items
 type ItemList struct {
+	// Results defines the total number of items in the collection matching the current
+	// context without any filter applied. If the storage handler cannot compute this value, -1 is set.
+	Results int
 	// Total defines the total number of items in the collection matching the current
 	// context. If the storage handler cannot compute this value, -1 is set.
 	Total int
 	// Page is the current page represented by this ItemList.
 	Page int
+	// Offset is the offset of the query (mongodb skip)
+	Offset int
 	// Items is the list of items contained in the current page given the current
 	// context.
 	Items []*Item
--- a/resource/storage.go
+++ b/resource/storage.go
@@ -20,7 +20,7 @@ type Storer interface {
 	// If the fetching of the data is not immediate, the method must listen for cancellation
 	// on the passed ctx. If the operation is stopped due to context cancellation, the
 	// function must return the result of the ctx.Err() method.
-	Find(ctx context.Context, lookup *Lookup, page, perPage int) (*ItemList, error)
+	Find(ctx context.Context, lookup *Lookup, page, perPage int, skip int) (*ItemList, error)
 	// Insert stores new items in the backend store. If any of the items does already exist,
 	// no item should be inserted and a resource.ErrConflict must be returned. The insertion
 	// of the items must be performed atomically. If more than one item is provided and the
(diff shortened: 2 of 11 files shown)


resource/item.go
```diff
@@ -24,11 +24,16 @@ type Item struct {
 
 // ItemList represents a list of items
 type ItemList struct {
+	// Results defines the total number of items in the collection matching the current
+	// context without any filter applied. If the storage handler cannot compute this value, -1 is set.
+	Results int
```
rs: This chunk is not part of offset support and should be removed/proposed as a separate PR (I'm not sure to understand the use-case though).

resource/storage.go
```diff
@@ -20,7 +20,7 @@ type Storer interface {
 	// If the fetching of the data is not immediate, the method must listen for cancellation
 	// on the passed ctx. If the operation is stopped due to context cancellation, the
 	// function must return the result of the ctx.Err() method.
-	Find(ctx context.Context, lookup *Lookup, page, perPage int) (*ItemList, error)
+	Find(ctx context.Context, lookup *Lookup, page, perPage int, skip int) (*ItemList, error)
```
rs: This is a breaking change. I'm not against it but we need to at least update all the supported storage handlers:

* https://github.com/rs/rest-layer-mem
* https://github.com/rs/rest-layer-mongo
* https://github.com/rs/rest-layer-es
* https://github.com/rs/rest-layer-hystrix

Alternatively we may think of an optional interface to introduce this change. But on the long run it would add complexity so a breaking but supported change is fine by me.

omani: my contributions will support the rest-layer-mongo and I would take rest-layer-mem into account. I can't support elasticsearch and hystrix though if this is ok for you.

resource/item.go
omani: yes it does not belong to this commit. this was a mistake. but I would like to keep it since I've also extended `ItemList`.

`Results` is basically the total count of items fetched by mgo with any query params like limit and/or offset applied. and total is now the total count of items without any filters applied.
if someone wants to get meta information with the response (see https://github.com/rs/rest-layer/#custom-response-formatter--sender) it looks like this:

"meta": {"offset": 2,"page": 1,"results": 1,"total": 4}

omani:
>my contributions will support rest-layer-mongo and I would take rest-layer-mem into account. I can't support elasticsearch and hystrix though if this is ok for you.

resource/item.go
rs: Please remove it from this PR and let's discuss it in a separate PR.
I'm not ok to change the meaning of `total`, and the distinction between `results` and `total` is not clear. My understanding is that `total` would become the total number of document in the collection, but this is something that would not apply to most backend storage engines. I'm not sure it is relevant to have both, it is more confusing that anything else.

Again, let's discuss that separately.

resource/storage.go
rs: I will do them, it's ok.

rs: BTW, why naming this new param `skip` instead of `offset` as it is named in the query-string and in Mongo?

omani: first I started with `skip` then decided to use `offset` since this is a better term when doing queries and people know what it is meant to be. but then I just changed the query param to offset and left `skip` in all functions as is.

I am just about to push the PR. do you want `skip` or `offset` as param in the funcs?

omani: I think `offset` as param name is better and we know what it refers to.

omani:
done.

rs:
Where? Why did you close the PR?

omani:
the PR is #62

PR #62, opened 2016-12-13
Provide: offset query param

omani:
as requested.

### DIFF
--- a/graphql/query.go
+++ b/graphql/query.go
@@ -81,41 +81,49 @@ func listParamResolver(r *resource.Resource, p graphql.ResolveParams, params url
 	offset = 1
 	// Default value on non HEAD request for limit is -1 (pagination disabled)
 	limit = -1
+
 	if l := r.Conf().PaginationDefaultLimit; l > 0 {
 		limit = l
 	}
-	if p, ok := p.Args["offset"].(string); ok && p != "" {
+	if s, ok := p.Args["skip"].(string); ok && s != "" {
+		i, err := strconv.ParseUint(s, 10, 32)
+		if err != nil {
+			return nil, 0, 0, errors.New("Invalid `skip` parameter")
+		}
+		offset = int(i)
+	}
+	if p, ok := p.Args["page"].(string); ok && p != "" {
 		i, err := strconv.ParseUint(p, 10, 32)
 		if err != nil {
-			return nil, 0, 0, errors.New("invalid `limit` parameter")
+			return nil, 0, 0, errors.New("Invalid `page` parameter")
 		}
 		offset = int(i)
 	}
 	if l, ok := p.Args["limit"].(string); ok && l != "" {
 		i, err := strconv.ParseUint(l, 10, 32)
 		if err != nil {
-			return nil, 0, 0, errors.New("invalid `limit` parameter")
+			return nil, 0, 0, errors.New("Invalid `limit` parameter")
 		}
 		limit = int(i)
 	}
-	if limit == -1 && offset != 1 {
-		return nil, 0, 0, errors.New("cannot use `offset' parameter with no `limit' paramter on a resource with no default pagination size")
+	if offset >= 0 && limit == -1 {
+		return nil, 0, 0, errors.New("cannot use `page' parameter with no `limit' paramter on a resource with no default pagination size")
 	}
 	lookup = resource.NewLookup()
 	if sort, ok := p.Args["sort"].(string); ok && sort != "" {
 		if err := lookup.SetSort(sort, r.Validator()); err != nil {
-			return nil, 0, 0, fmt.Errorf("invalid `sort` parameter: %v", err)
+			return nil, 0, 0, fmt.Errorf("Invalid `sort` parameter: %v", err)
 		}
 	}
 	if filter, ok := p.Args["filter"].(string); ok && filter != "" {
 		if err := lookup.AddFilter(filter, r.Validator()); err != nil {
-			return nil, 0, 0, fmt.Errorf("invalid `filter` parameter: %v", err)
+			return nil, 0, 0, fmt.Errorf("Invalid `filter` parameter: %v", err)
 		}
 	}
 	if params != nil {
 		if filter := params.Get("filter"); filter != "" {
 			if err := lookup.AddFilter(filter, r.Validator()); err != nil {
-				return nil, 0, 0, fmt.Errorf("invalid `filter` parameter: %v", err)
+				return nil, 0, 0, fmt.Errorf("Invalid `filter` parameter: %v", err)
 			}
 		}
 	}
--- a/graphql/query.go
+++ b/graphql/query.go
@@ -78,7 +78,7 @@ var listArgs = graphql.FieldConfigArgument{
 }
 
 func listParamResolver(r *resource.Resource, p graphql.ResolveParams, params url.Values) (lookup *resource.Lookup, offset int, limit int, err error) {
-	offset = 1
+	offset = 0
 	// Default value on non HEAD request for limit is -1 (pagination disabled)
 	limit = -1
 
--- a/graphql/query.go
+++ b/graphql/query.go
@@ -63,7 +63,7 @@ func (t types) getGetQuery(idx resource.Index, r *resource.Resource) *graphql.Fi
 }
 
 var listArgs = graphql.FieldConfigArgument{
-	"page": &graphql.ArgumentConfig{
+	"offset": &graphql.ArgumentConfig{
 		Type: graphql.Int,
 	},
 	"limit": &graphql.ArgumentConfig{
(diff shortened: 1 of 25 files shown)


resource/storage.go
```diff
@@ -155,7 +155,7 @@ func (s storageWrapper) MultiGet(ctx context.Context, ids []interface{}) (items
 }
 
 // Find tries to use storer MultiGet with some pattern or Find otherwise
-func (s storageWrapper) Find(ctx context.Context, lookup *Lookup, page, perPage int) (list *ItemList, err error) {
+func (s storageWrapper) Find(ctx context.Context, lookup *Lookup, page, perPage, offset int) (list *ItemList, err error) {
```
smyrman: If adding `offset`, can `page` actually be removed, and then let the REST API code supply offset = page*perPage when page is specified in the request? Or is there arreason that won't work?

It's not very clean code to supply more parameters than needed...

Also, if replacing page with offset, is perPage better renamed to limit to avoid confusion?

The comment applies to all related changes.

omani: `offset` can't be used in combination with `page`. I've also said this in the docs. so `page` is still needed. in fact, `page` is always 1 anyway when `offset` is used. so `pagePage` would still be descriptive even then.

smyrman:
@omani, you know you can update a PR instead of closing it bu pushing and/or force-pushing to the same branch. Having to jump to a new PR all the time makes the review comments hard to track!

omani:
yes I know. the reason I closed the former PR was because I wanted to start a clean history (and I messed up my commits and had to restart again).

smyrman:
> yes I know. the reason I closed the former PR was because I wanted to start a clean history (and I messed up my commits and had to restart again).

You can do both rebase and force-push before the final approval. That's common to do anyway to clean up the history.

resource/storage.go
omani: I think I understand what you mean. the math is the same. but a client requesting with `page` has a different intention behind it as one who wants an offset.

so, while:
```
Find(ctx, query, 2, 5, 0) // sets page to 2
Find(ctx, query, 0, 5, 5*(2-1)) // sets offset instead of page
```

may be the same. this isn't:
```
curl http://localhost:8080/users?offset=10&limit=10
curl http://localhost:8080/users?page=2&limit=10
```

smyrman:
```
curl http://localhost:8080/users?offset=10&limit=10
curl http://localhost:8080/users?page=2&limit=10
```
Could you please elaborate on why these are not equivalent? I have used the first option here to implement pagination in other APIs. As for the backend, why does it need to care weather page or offset was specified by the user, if it is to return the same results anyway?

smyrman: > Could you please elaborate on why these are not equivalent?

I do understand that limit gives the end-user more options than page, and that page is slightly easier to use for pagination than limt. I am not arguing for removing page as a REST-API parameter.

omani: remember I am not talking about the math behind it. because then:
```
curl http://localhost:8080/users?offset=10&limit=10
curl http://localhost:8080/users?page=2&limit=10
```
would be the same of course. I am talking about the intentions a client has by describing it in  the query params.

smyrman: > I hope I could explain it.

Maybe we are talking past each other than, and maybe I replied a bit quickly to your previous comment. I do agree on the semantic differences for the REST API point of view.

However, would you agree that these semantic differences are irrelevant for the Storage inteface / storage backends, which is what I am suggesting to simplify here?
>

omani: well. yes I agree. for the storage backend you could do the translation in [rest/method_get.go](https://github.com/rs/rest-layer/pull/62/commits/ca55f79fc1e83a989ea5912c470c358d3ad02cf6#diff-a90cd58ba7082f803eb1ab4b75bd612a) and avoid one more func param.

omani: since I am quite new to this project I would let you both or @rs in case decide.

rs: Why could not we use page and offset at the same time. What of I want to paginate with 10 item per page but shifted by 5 items at the begining?

omani: I can't think of any use case for shifting by N items and then have a pagination available for the result set.

I also can't think of any use case for shifting by N items each page iteration on the result set.

It's still doable though.

smyrman: @rs, even if offset and page was allowed at the same time, which I don't have any opinion for or against, I still think that's a detail that is irrelevant for the Storage backend. For the Storage backend we should make the interface simple. Would you agree?

smyrman: Also it makes the implementation easier to change later if the Storage engine Find just expects offset, as the logic for accepting both page and offset can be changed in the one file @Omani mentions.

rs: Yes you're right, the storage should only take `offset` and `limit` and the API should translate `page`, `perPage` and `skip` parameters into `offset`/`limit`.

omani:
updated.

graphql/query.go
```diff
@@ -77,29 +77,29 @@ var listArgs = graphql.FieldConfigArgument{
 	},
 }
 
-func listParamResolver(r *resource.Resource, p graphql.ResolveParams, params url.Values) (lookup *resource.Lookup, page int, perPage int, err error) {
-	page = 1
-	// Default value on non HEAD request for perPage is -1 (pagination disabled)
-	perPage = -1
+func listParamResolver(r *resource.Resource, p graphql.ResolveParams, params url.Values) (lookup *resource.Lookup, offset int, limit int, err error) {
+	offset = 1
```
smyrman: Should not the default offset be 0?

graphql/query.go
```diff
@@ -77,29 +77,29 @@ var listArgs = graphql.FieldConfigArgument{
...
 		i, err := strconv.ParseUint(l, 10, 32)
 		if err != nil {
 			return nil, 0, 0, errors.New("invalid `limit` parameter")
 		}
-		perPage = int(i)
+		limit = int(i)
 	}
-	if perPage == -1 && page != 1 {
-		return nil, 0, 0, errors.New("cannot use `page' parameter with no `limit' paramter on a resource with no default pagination size")
+	if limit == -1 && offset != 1 {
```
smyrman: I assume it should be `offset != 0`?

graphql/query.go
rs: Yes I think so too.

resource/item.go
```diff
@@ -29,6 +29,8 @@ type ItemList struct {
 	Total int
 	// Page is the current page represented by this ItemList.
 	Page int
+	// Offset is the offset of the query (mongodb skip)
```
rs: Do not forget the comment.

rest/method_get.go
```diff
@@ -58,6 +59,20 @@ func listGet(ctx context.Context, r *http.Request, route *RouteMatch) (status in
 		e = NewError(err)
 		return e.Code, nil, e
 	}
+	// Now we can set Total.
+	list.Total = len(list.Items)
```
rs: This is the current page size, not the total. Only the storage handler can know (or not) the total elements in the collection.

rest/method_get.go
```diff
@@ -58,6 +59,20 @@ func listGet(ctx context.Context, r *http.Request, route *RouteMatch) (status in
...
+
+	// Set appropriate fields.
+	if page >= 0 {
+		list.Page = page + 1
+		list.Skip = ((page + 1) * limit) - limit
```
rs: `Skip` should be equal to the `skip` parameter. What are you calculating here? I don't follow.

rs:
I made required changes and merged to master.

rest/method_get.go
omani: you are right. I was on the wrong path here. But why can't we just say:
```go
total, err := c.Count()
if err != nil {
	return nil, err
}
...
list := &resource.ItemList{Total: total, Items: []*resource.Item{}}
...
```

to get the total items at least in rest-layer-mongo? Is it because of performance considerations?

rs: Yes, getting the total count can be arbitrary complex for the storage and it is an information that is not fomally requested and most of the time not used.

We may define a way to formally ask for this info, and thus force the computation explicitly.

omani: >it is an information that is not fomally requested and most of the time not used.

I don't agree to this. Even when providing pagination it makes sense to provide this information also.

Imagine a frontend UI fetching items from db and displaying it in a table with page numbers. without total items count, how will you know how many page numbers to display?

rs: That is one use case. Your UI may choose not to show the total number of pages or just an estimate (ala Google) or you may use the API for some other purpose. I'm not saying it is not useful, I'm just saying it is not always useful, and I don't think it is wise to have the consumer of the API paying for the computation of this information when not explicitly requested.

Issue #63, opened 2016-12-15
[jsonschema] Bug: error not returned from Encoder


smyrman:
While donig some test-cleanup in the `jsonschema` package, I found a bug where errors in encoding a validator are never returned from `jsonschema.Encoder`.

I found this bug while changing this test:
```go
-func TestErrNotImplemented(t *testing.T) {
-       validator := &schema.IP{}
-       b := new(bytes.Buffer)
-       assert.Equal(t, ErrNotImplemented, validatorToJSONSchema(b, validator))
-}
```
Into one that tests the public interface:
```go
+type dummyValidator struct{}
+
+func (v dummyValidator) Validate(value interface{}) (interface{}, error) {
+       return value, nil
+}
+
+func TestErrNotImplemented(t *testing.T) {
+       s := schema.Schema{
+               Fields: schema.Fields{
+                       "i": {
+                               Validator: &dummyValidator{},
+                       },
+               },
+       }
+       enc := jsonschema.NewEncoder(new(bytes.Buffer))
+       assert.Equal(t, jsonschema.ErrNotImplemented, enc.Encode(&s))
+}
```

PS! The result is the same if  `&schema.IP{}` is used as `Validator` in the new test.

I am trying to see if I can fix this together with my change.

smyrman:
@yanfali, pinging you about this so you are aware. Should be fixed by #64.

smyrman:
Since #64 is merged, this issue can be closed.

PR #64, opened 2016-12-15
jsonschema: Test cleanup and bugfix

smyrman:
- Replaces encoder_test.go (encoding package) with example_test.go (jsonschema package).
- Replaces remaining tests in jsonschema_test.go with (table-)tests
  testing the pulbic interface. The new tests are placed in
  {typeName}_test.go or all_test.go.
- Solves Issue #63 (found by test change) where errors from
  validatorToJSONSchema and serializeField where never propagated.


smyrman:
I should mention, this change makes all the tests go1.7+, as I see the CI no longer runs Go 1.6

smyrman:
> The new tests are placed in {typeName}_test.go or all_test.go.

FYI: #51 will aim to move encoder logic into  {typeName}.go as well, which will match nicely with this.

schema/encoding/jsonschema/example_test.go
```diff
@@ -0,0 +1,59 @@
...
+	// | | | "readOnly": true,
+	// | | | "type": "string",
+	// | | | "maxLength": 42
+	// | | },
+	// | | "foobar": {}
+	// | },
+	// | "required": [
+	// | | "foo"
+	// | ]
+	// }
```
smyrman: This output does not always seem to match! Please don't merge until fixed.

smyrman:
Ready from my point of view now

schema/encoding/jsonschema/jsonschema.go
```diff
@@ -74,6 +73,18 @@ func (ew errWriter) writeBytes(b []byte) {
...
 }
 
+// sortedFieldNames returns a list with all field names alphabetically sorted.
+func sortedFieldNames(v schema.Fields) []string {
+	keys := make([]string, len(v))
+	i := 0
+	for k := range v {
+		keys[i] = k
+		i++
+	}
```
rs: I think it is more idiomatic (and equivalent perf-wise) to do:

```go
	keys := make([]string, 0, len(v))
	for k := range v {
		keys = append(keys, k)
	}
```

smyrman: OK, I will change it shortly.

smyrman: Done

Issue #68, opened 2016-12-22
Feature request: have total items in response


omani:
Hi,

as previously discussed in a PR. I would like to ask how and when do we implement the feature of having `Total` items being displayed in the response? @rs said it should be explicitly requested, so my question is, are there any plans how to do it?

rs:
We have to define the way we request it and how to query the storage layer.

I would propose a query-string parameter like `total=1` (no sure about the naming though) and an optional interface on storer with the method `Count`. If `total` is `1`, the storer implement this `Counter` interface and the returned `Total` is -1, rest-layer would then call the `Count` method to fill this metadata in the response.

cc @smyrman

smyrman:
@rs, I think that proposal is reasonable, but I will list one more.

If it is not necessary to decide this at request time, I.e. the `total=1/0` rest parametre is not needed, I also have an alternative solution:

Backends that don't have an efficient way of doing the total count together with the query, could offer an additional Storer implementation with the suffix `WithTotal`. If this is not done, the end-user could implament his own `Storer` wrapper (e.get. using struct inheritance) that does am additional count queryour as part of `Find()`.

smyrman:
So, basically it boils down to weather includingelin the `total` count should be a server side or client side decision... Are there cases where an additional count query, e.g. for large MySQL tables, could lead to server side lockups or worse?

There could also be that there are other good usecase for the `Counter` interface to be added, unrelated to this feature. I don't foresee any at this point, but it's possible...

rs:
I don't understand the different between your proposal and mine @smyrman. Did I miss something?

smyrman:
@rs ,  the alternative proposal does not include adding a new optional interface.

rs:
When the database is not able to compute the total efficiently at query time, you can't predict the complexity of this query. When filters are involved and indexes don't cover the whole query, counting may be more expensive than the *paged* query itself.

So my goal is to offer possibility to the client to only request this total when it is wanted.

smyrman:
> So my goal is to offer possibility to the client to only request this total when it is wanted.

If the clients should make this call, it's fine, then the alternative one don't work. My angle was (potentially third-party) clients forcing the server to a hault by generating count queries that never complete, and possibly don't get killed when the request times out. In this case it sounds reasonable to be able to disable total=1 server side. With your proposal, I was _picturing_ that all supported storage engines would implement the optional Counter interface, but of course that does not need to be the case here.

rs:
[...]

PR #69, opened 2016-12-22
Allow user (or config) to explicitely request the list total metadata

rs:
Fix: #68

### DIFF
--- a/resource/conf.go
+++ b/resource/conf.go
@@ -8,7 +8,35 @@ type Conf struct {
 	// no default page size is set resulting in no pagination if no `limit` parameter
 	// is provided.
 	PaginationDefaultLimit int
+	// ForceTotal controls how total number of items on list request is computed.
+	// By default (TotalOptIn), if the total cannot be computed by the storage
+	// handler for free, no total metadata is returned until the user explicitely
+	// request it using the total=1 query-string parameter. Note that if the
+	// storage cannot compute the total and does not implement the resource.Counter
+	// interface, a "not implemented" error is returned.
+	//
+	// The TotalAlways mode always force the computation of the total (make sure the
+	// storage either compute the total on Find or implement the resource.Counter
+	// interface.
+	//
+	// TotalDenied prevents the user from requesting the total.
+	ForceTotal ForceTotalMode
 }
+
+// ForceTotalMode defines Conf.ForceTotal modes
+type ForceTotalMode int
+
+const (
+	// TotalOptIn allows the end-user to opt-in to forcing the total count by
+	// adding the total=1 query-string parameter.
+	TotalOptIn ForceTotalMode = iota
+	// TotalAlways always force the total number of items on list requests
+	TotalAlways
+	// TotalDenied disallows forcing of the total count, and returns an error
+	// if total=1 is supplied, and the total count is not provided by the
+	// Storer's Find method.
+	TotalDenied
+)
 
 // Mode defines CRUDL modes to be used with Conf.AllowedModes.
 type Mode int
--- a/README.md
+++ b/README.md
@@ -655,6 +655,7 @@ The `resource.Conf` type has the following customizable properties:
 | ------------------------ | -------------
 | `AllowedModes`           | A list of `resource.Mode` allowed for the resource.
 | `PaginationDefaultLimit` | If set, pagination is enabled by default with a number of item per page defined here.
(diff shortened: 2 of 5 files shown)


resource/conf.go
```diff
@@ -8,8 +8,35 @@ type Conf struct {
 	// no default page size is set resulting in no pagination if no `limit` parameter
 	// is provided.
 	PaginationDefaultLimit int
+	// ForceTotal controls how total number of items on list request is computed.
+	// By default (User), if the total cannot be computed by the storage andler for
```
smyrman: Typo andler=handler

resource/conf.go
```diff
@@ -8,8 +8,35 @@ type Conf struct {
...
+	ForceTotal ForceTotalMode
 }
 
+// ForceTotalMode defines Conf.ForceTotal modes
+type ForceTotalMode int
+
+const (
+	// User let the end-user request the total number of items on list requests
+	// using the total=1 query-string parameter
+	User ForceTotalMode = iota
```
smyrman: Is there a better name for the default state than `User`?

resource/conf.go
```diff
@@ -8,8 +8,35 @@ type Conf struct {
...
+	// Always always force the total number of items on list requests
+	Always
```
smyrman: Should this state also allow `total=0` to opt-out of the default? Or is that silly? If it should, than `Always` would be misleading as well.

smyrman: What about?
```go
const (
       // OptIn allows the user to opt-in to forcing the total count by adding `total=1`.
        OptIn ForceTotalMode = iota
       // OptOut allows the user to opt-out of forcing the total count by adding `total=0`.
        OptOut
        // Denied disallows forcing of the total count, and returns an error if total=1 is
        // supplied, _and_ the total count is not provided by the Storer's Find method.
        Denied
)
```

resource/conf.go
```diff
@@ -8,8 +8,35 @@ type Conf struct {
...
+	//
+	// The Always mode always force the computation of the total (make sure the
+	// storage either compute the total on Find or implement the resource.Counter
+	// interface.
+	//
+	// Denied prevents the user from requesting the total.
+	ForceTotal ForceTotalMode
 }
 
+// ForceTotalMode defines Conf.ForceTotal modes
```
smyrman: Adding a trailing period, also for single-sentence comments, would be [more Idiomatic](https://golang.org/doc/effective_go.html#commentary), and in line with how the standard library [do single-sentence comments](https://golang.org/pkg/io/#ByteWriter). This is a small detail, so feel free to ignore it if you prefer this style.


The concrete text I am referring to from Effective Go is:

 > Doc comments work best as complete sentences, which allow a wide variety of automated presentations. The first sentence should be a one-sentence summary that starts with the name being declared.

P.S! Last time I did a PR for rest-layer, I chose not to add the trailing period to stay consistent with other comments in the same file.

resource/conf.go
rs: Would work, naming is hard. I just can't see a use-case where the user would like to opt-out, do you see one?

resource/conf.go
rs: I agree. I would prefer to fix all comments in a separate PR. Do you know some refact tools to help us with that?

smyrman: I don't know any for this specific purpose.

resource/conf.go
smyrman: > I just can't see a use-case where the user would like to opt-out, do you see one?

You are right, I also fail to see any _strong_ use-case for it. I see a _weak_ one, but let's not worry about that.

Still, it would be nice to rename `User` to avoid confusion and perhaps also add a `Mode` prefix (ala `http.Status...`). I was quite confused the first time I read "By default (User)". It's not completely obvious what it means... Also remember that it will appear as `resource.User` in auto-completion and code, which could even lead people to believe it's a user object.

Would `ModeOptIn`, `ModeAlways`, `ModeDenied` be better? This even allows `ModeOptOut` to be added later if anyone actually requests it.

rs: Good point.

omani:
question: I see a total in my response, without changing/setting any configuration. is this expected?

resource/conf.go
omani: this should be doable with unix tools on cli.

omani:
ok I see why. it is the from line 236 in https://github.com/rs/rest-layer-mongo/commit/a96dfce9b8e108cdbaca48260bb0870cea650b8a.

PR #70, opened 2016-12-22
jsonschema: #51 - split into multiple files

smyrman:
This change splits the encoder logic into a separate file per supported
schema.FieldValidator type. It also provides some minor clenaups to the
documentation and adds some FIXME statements.

Functions / type defintions are declared in the order in which they are
called, which should make the code "read like an article from top to
bottom".

This cleanup provides the second step of #51.


smyrman:
CC @yanfali and @rs

smyrman:
I think the failures here are related to #69 not being merged.

smyrman:
Chnges has been rebased.

rs:
Is it ready to merge?

smyrman:
Yes

PR #74, opened 2016-12-27
Provide regex query param and regex filter

### LINKED ISSUE #73: Provide: mongo's $regex query parameter
It would be awesome to be able to query a value of a field by regex.

Since the format of the parameter is inspired by the mongodb query format we could also have its $regex in queries.

omani:
fixes #73

### DIFF
--- a/README.md
+++ b/README.md
@@ -823,13 +823,24 @@ The opposite `$nin` is also available.
 
 The following numeric comparisons operators are supported: `$lt`, `$lte`, `$gt`, `$gte`.
 
-The `$exists` operator matches document containing the field, even if this field is `null`
+The `$exists` operator matches documents containing the field, even if this field is `null`
 
 ```json
 {"type": {"$exists": true}}
 ```
 
 You can invert the operator by passing `false`.
+
+There is also a `$regex` operator that matches documents containing the field given as a regular expression. The syntax of the regular expressions accepted is the same general syntax used by Perl, Python, and other languages.
+More precisely, it is the syntax accepted by RE2 and described at https://golang.org/s/re2syntax, except for \C.
+
+For example the following regular expression would match any document with a field `type` and its value `rest-layer`.
+
+```json
+{"type": {"$regex": "re[s]{1}t-la.+r"}}
+```
+
+However, keep in mind that Storers have to support regular expression. An error of `ErrNotImplemented` will be returned for those storage backends which do not support the `$regex` operator.
 
 ### Filter operators
 
@@ -843,6 +854,7 @@ You can invert the operator by passing `false`.
 | `$gt`     | `{"a": {"$gt": 10}}`                | Fields value is greater than specified number.
 | `$gte`    | `{"a": {"$gte": 10}}`               | Fields value is greater than or equal to the specified number.
 | `$exists` | `{"a": {"$exists": true}}`          | Match if the field is present (or not if set to `false`) in the item, event if `nil`.
+| `$regex`  | `{"a": {"$regex": "fo[o]{1}"}}`     | Match regular expression on a field's value.
 
 *Some storage handlers may not support all operators. Refer to the storage handler's documentation for more info.*
 
--- a/README.md
+++ b/README.md
@@ -834,10 +834,23 @@ You can invert the operator by passing `false`.
 There is also a `$regex` operator that matches documents containing the field given as a regular expression. In general, the syntax of the regular expressions accepted is the same general syntax used by Perl, Python, and other languages.
 More precisely, it is the syntax accepted by RE2 and described at https://golang.org/s/re2syntax, except for \C.
 
+Flags are supported for more control over regular expressions. Flag syntax is xyz (set) or -xyz (clear) or xy-z (set xy, clear z).
+The flags are:
+i              case-insensitive (default false)
+m              multi-line mode: ^ and $ match begin/end line in addition to begin/end text (default false)
+s              let . match \n (default false)
+U              ungreedy: swap meaning of x* and x*?, x+ and x+?, etc (default false)
+
 For example the following regular expression would match any document with a field `type` and its value `rest-layer`.
 
 ```json
 {"type": {"$regex": "re[s]{1}t-la.+r"}}
+```
+
+The same example with flags:
+
+```json
+{"type": {"$regex": "(?i)re[s]{1}t-LAYER"}}
 ```
 
 However, keep in mind that Storers have to support regular expression and depending on the implementation of the storage handler the accepted syntax may vary.
--- a/README.md
+++ b/README.md
@@ -831,7 +831,7 @@ The `$exists` operator matches documents containing the field, even if this fiel
 
 You can invert the operator by passing `false`.
 
-There is also a `$regex` operator that matches documents containing the field given as a regular expression. The syntax of the regular expressions accepted is the same general syntax used by Perl, Python, and other languages.
+There is also a `$regex` operator that matches documents containing the field given as a regular expression. In general, the syntax of the regular expressions accepted is the same general syntax used by Perl, Python, and other languages.
 More precisely, it is the syntax accepted by RE2 and described at https://golang.org/s/re2syntax, except for \C.
 
 For example the following regular expression would match any document with a field `type` and its value `rest-layer`.
@@ -840,7 +840,8 @@ For example the following regular expression would match any document with a fie
 {"type": {"$regex": "re[s]{1}t-la.+r"}}
 ```
 
-However, keep in mind that Storers have to support regular expression. An error of `ErrNotImplemented` will be returned for those storage backends which do not support the `$regex` operator.
+However, keep in mind that Storers have to support regular expression and depending on the implementation of the storage handler the accepted syntax may vary.
+An error of `ErrNotImplemented` will be returned for those storage backends which do not support the `$regex` operator.
 
 ### Filter operators
 
--- a/schema/query.go
+++ b/schema/query.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"regexp"
 )
 
 // Query defines an expression against a schema to perform a match on schema's data.
@@ -84,6 +85,12 @@ type LowerOrEqual struct {
 	Value float64
 }
 
+// Regex matches values that match to a specified regular expression.
+type Regex struct {
+	Field string
+	Value string
+}
+
 // NewQuery returns a new query with the provided key/value validated against validator
 func NewQuery(q map[string]interface{}, validator Validator) (Query, error) {
 	return validateQuery(q, validator, "")
@@ -107,6 +114,17 @@ func validateQuery(q map[string]interface{}, validator Validator, parentKey stri
 	queries := Query{}
 	for key, exp := range q {
 		switch key {
+		case "$regex":
+			if parentKey == "" {
+				return nil, errors.New("$regex can't be at first level")
+			}
+			regex, ok := exp.(string)
+			if !ok {
+				return nil, errors.New("$regex can only get strings as value")
+			}
+			if regex != "" {
+				queries = append(queries, Regex{Field: parentKey, Value: regex})
+			}
 		case "$exists":
 			if parentKey == "" {
 				return nil, errors.New("$exists can't be at first level")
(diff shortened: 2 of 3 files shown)


omani:
rest-layer-mongo implements this in https://github.com/rs/rest-layer-mongo/pull/7

schema/query.go
```diff
@@ -356,3 +374,9 @@ func (e LowerOrEqual) Match(payload map[string]interface{}) bool {
 	n, ok := isNumber(getField(payload, e.Field))
 	return ok && (n <= e.Value)
 }
+
+// Match implements Expression interface
+func (e Regex) Match(payload map[string]interface{}) bool {
+	var regexpedValue = regexp.MustCompile(e.Value)
```
rs: You don't want to compile the regexp at runtime. The `Regex` type should store the compiled version of the regexp. This would also provide init time validation of the regexp.

omani: I pushed the changes.

schema/query.go
```diff
@@ -123,7 +123,7 @@ func validateQuery(q map[string]interface{}, validator Validator, parentKey stri
 				return nil, errors.New("$regex can only get strings as value")
 			}
 			if regex != "" {
-				queries = append(queries, Regex{Field: parentKey, Value: regex})
+				queries = append(queries, Regex{Field: parentKey, Value: regexp.MustCompile(regex)})
```
rs: Please use `Compile` instead and return the error if any. `MustCompile` would make the program to panic without much explanation.

omani: yes. I totally forgot. I am just about to push my fix.

schema/query.go
```diff
@@ -107,6 +114,21 @@ func validateQuery(q map[string]interface{}, validator Validator, parentKey stri
...
+				return nil, errors.New("$regex can't be at first level")
+			}
+			regex, ok := exp.(string)
+			if !ok {
+				return nil, errors.New("$regex can only get strings as value")
+			}
+			if regex != "" {
+				v, err := regexp.Compile(regex)
+				if err != nil {
+					return nil, err
```
rs: `return nil, fmt.Errorf("$regex: invalid regex: %v", err)`

schema/query_test.go
```diff
@@ -193,6 +198,8 @@ func TestParseQueryInvalidHierarchy(t *testing.T) {
 	assert.EqualError(t, err, "$gt can't be at first level")
 	_, err = ParseQuery("{\"$in\": [1,2]}", s)
 	assert.EqualError(t, err, "$in can't be at first level")
+	_, err = ParseQuery("{\"$regex\": \"someregexpression\"}", s)
+	assert.EqualError(t, err, "$regex can't be at first level")
```
rs: Please test some invalid expression.

omani: will do.

rs:
One thing concerns me now: we compile the regexp unconditionally to catch errors early and being able to report them. But in most cases, we won't use the compiled version of the regexp as the storage handler will pass the string version to the engine. Only the mem engine and the dependency checker are currently using the rest-layer provided matcher.

This will thus have a performance impact on `$regex` queries, but I don't see how to do it otherwise. One solution could be to use a LRU cache to cache compiled regexps, as most of the time, identical regexp are used in production, but it may not be a wanted solution in all use-cases.

Any thoughts?

schema/query_test.go
omani: done.

rs:
@omani: do not forget to update the doc: http://rest-layer.io/#filtering

rs:
We should also think about a way to support regex options.

omani:
updated.

README.md
```diff
@@ -823,14 +823,25 @@ The opposite `$nin` is also available.
...
+The `$exists` operator matches documents containing the field, even if this field is `null`
 
 ```json
 {"type": {"$exists": true}}
 ```
 
 You can invert the operator by passing `false`.
 
+There is also a `$regex` operator that matches documents containing the field given as a regular expression. The syntax of the regular expressions accepted is the same general syntax used by Perl, Python, and other languages.
+More precisely, it is the syntax accepted by RE2 and described at https://golang.org/s/re2syntax, except for \C.
```
rs: I think we should say that the accepted syntax may vary depending on the storage handler.

omani:
the comment was deleted by you. but I read it and committed.

schema/query.go
omani: fixed.

smyrman:
> One thing concerns me now: we compile the regexp unconditionally to catch errors early and being able to report them. But in most cases, we won't use the compiled version of the regexp as the storage handler will pass the string version to the engine.

Would it be practical to just let the backend validate the regex on it's own? That would also be more efficient when a storage backend only supports partial regex...

rs:
That would be ideal but then the `Match()` method would have no way to report invalid regexp (this was the initial state of this PR).

omani:
are there any other things I need to do here?

rs:
Yes, regexp option handling.

Also, add a TOFIX comment on the `Compile` so we keep in mind that we have to find a better solution for this.

omani:
can you explain what you mean by regex option handling? an example would be appreciated.

rs:
Check the `$options` param on mongo: https://docs.mongodb.com/manual/reference/operator/query/regex/.

omani:
ok I will implement this. but it looks like only mongo will be able to know what to do with $options. so I guess any other storage handler will return an ErrNotImplemented if $options query param is passed or will just ignore it?

rs:
I'm not sure `$options` is the best solution. Maybe we could support `$regexp` with either a scalar or a an array: `{"$regex": ["som.\sregexp?", "i"]}`.

Most engine supporting regex query will support options. ElasticSearch calls it "flags" for instance.

smyrman:
> Most engine supporting regex query will support options. ElasticSearch calls it "flags" for instance.

Go itself also supports [flags](https://golang.org/pkg/regexp/syntax/#Flags), e.g. through using [Parse()](https://golang.org/pkg/regexp/syntax/#Parse) from `regexp` syntax.

However the supported options by MongoDB, Go, Elasticsearch and other potential backends, such as [PostgreSQL](https://www.postgresql.org/docs/current/static/functions-matching.html#FUNCTIONS-POSIX-TABLE), can vary somewhat.

To achieve more consistent behaviour between storage backends, would it make sense to do the `$options` support as a separate PR, giving us time to more carefully consider which options to support?

MongoDB also supports the syntax: `{ <field>: { $regex: /pattern/<options> } }`. Would that be more user-friendly than a separate URL parameter?

omani:
>MongoDB also supports the syntax: { <field>: { $regex: /pattern/<options> } }. Would that be more user-friendly than a separate URL parameter?

this makes parsing $regex more difficult. you would have to parse based on the given string and slashes are part of the regex.

>Maybe we could support $regexp with either a scalar or a an array: {"$regex": ["som.\sregexp?", "i"]}.
I like this approach. passing the value of $regex stays the same in `rest-layer` and the storage handler can differentiate between scalar/array and does the appropriate job.

omani:
I would implement my last comment.

omani:
I've implemented options handling. But before I commit I would like to ask how we deal with the regexp options in golang? shall I implement it also?

rs:
Go is prefixing the flags to the pattern using `(?flags)` notation. The problem is that this notation is not necessarily supported by all engines. Apparently MongoDB does when used with `$regex` and a pattern as string but I find no indication that ElasticSearch does for instance.

So we could decide to go for this syntax and extract those flags in storage handlers that do not support it.

omani:
@rs. yes just the way I thought this, too. my solution to this would be to check if options contains one of i,s,m,U and use it as (?flag). the rest goes without (?flags) in `Compile()`

omani:
what I mean is:
```go
case "$regex":
			if parentKey == "" {
				return nil, errors.New("$regex can't be at first level")
			}

			switch exp.(type) {
			case string:
				regex := exp.(string)
				v, err := regexp.Compile(regex)
				if err != nil {
					return nil, fmt.Errorf("$regex: invalid regex: %v", err)
				}

				value := RegexValue{
					RegValue:   v,
					RegOptions: "",
				}

				queries = append(queries, Regex{Field: parentKey, Value: value})
			case []interface{}:
				regex := exp.([]interface{})

				var (
					v   *regexp.Regexp
					err error
				)
				// Control regex options behavour.
				// First check if we find any options of the regexp package
				// Supported options (flags):
				// i              case-insensitive (default false)
				// m              multi-line mode: ^ and $ match begin/end line in addition to begin/end text (default false)
				// s              let . match \n (default false)
				// U              ungreedy: swap meaning of x* and x*?, x+ and x+?, etc (default false)
				if strings.ContainsAny(regex[1].(string), "imsU") {
					v, err = regexp.Compile("(?" + regex[1].(string) + ")" + regex[0].(string))
				} else {
					v, err = regexp.Compile(regex[0].(string))
				}
				if err != nil {
					return nil, fmt.Errorf("$regex: invalid regex: %v", err)
				}

				value := RegexValue{
					RegValue:   v,
					RegOptions: regex[1].(string),
				}
				queries = append(queries, Regex{Field: parentKey, Value: value})
			default:
				return nil, errors.New("$regex can only get string or array as value")
			}
```

the tests do pass. but now there is the problem of mixing flags of golang and mongo's. like:
```bash
http 'http://localhost:8080/hosts?access_token=eyJhbGc...' filter=='{"hostname": {"$regex": ["^FOO.+$", "ix"]}}'
```

here, mongo would know how to deal with the option "i" and "x". but since golang's regexp package does not have "x" I get:

```bash
"message": "Invalid `filter` parameter: $regex: invalid regex: error parsing regexp: invalid or unsupported Perl syntax: `(?ix`"
```

omani:
the above is because of `ContainsAny()` and to fix this is easy. I just want to know if this goes the right direction. feedback?

rs:
Do we really want to support Mongo's "x" flag? I don't think it is useful in the context of a REST API.

omani:
comitted.

schema/query.go
```diff
@@ -107,6 +121,56 @@ func validateQuery(q map[string]interface{}, validator Validator, parentKey stri
...
+					err error
+				)
+				// Control regex options behavour.
+				// First check if we find any options of the regexp package
+				// Supported options (flags):
+				// i              case-insensitive (default false)
+				// m              multi-line mode: ^ and $ match begin/end line in addition to begin/end text (default false)
+				// s              let . match \n (default false)
+				// U              ungreedy: swap meaning of x* and x*?, x+ and x+?, etc (default false)
+				if strings.ContainsAny(regex[1].(string), "imsU") {
```
smyrman: As it stands, this line may panic with index out of range. Also the .(string) typecast may result in a panic when ok is not checked.

I suggest you extract regex[0] and optionally regex[1] into two local variables while handling errors appropriately.

Other than that I assume ContainsAny won't check if regex[1] contains any other strings than imsU, so if I where to supply "imsUxab", would the regex.Comile then fail? Would perhaps be good to have a unittest for this:-)

schema/query_test.go
```diff
@@ -242,6 +264,10 @@ func TestQueryMatch(t *testing.T) {
 	assert.False(t, q.Match(map[string]interface{}{"foo": "bar"}))
 	assert.False(t, q.Match(map[string]interface{}{"bar": float64(1)}))
 	assert.True(t, q.Match(map[string]interface{}{"foo": "bar", "bar": float64(1)}))
+	q, _ = ParseQuery("{\"foo\": {\"$regex\": \"rege[x]{1}.+some\"}}", s)
+	assert.True(t, q.Match(map[string]interface{}{"foo": "regex-is-awesome"}))
+	q, _ = ParseQuery("{\"foo\": {\"$regex\": [\"^my.+-rest.+$\", \"iasdm\"]}}", s)
```
smyrman: should this actually result in an error?

schema/query.go
rs: I'm wondering if we need this notation after all. What about sticking to Go notation (i.e. embedded `(?flags)` prefix) and see later if we want to introduce more advanced notation.

In its current form, the code does not prevent you from using `(?flags)` with or without the array value format. So either we deal with this case or we just drop the array notation and just adopt Go notation.

Sorry @omani, I realize that I'm just telling you that your previous implementation was actually good enough :)

omani:
mongo counterpart https://github.com/rs/rest-layer-mongo/pull/7

README.md
```diff
@@ -823,14 +823,26 @@ The opposite `$nin` is also available.
...
+
+For example the following regular expression would match any document with a field `type` and its value `rest-layer`.
+
+```json
+{"type": {"$regex": "re[s]{1}t-la.+r"}}
+```
+
+However, keep in mind that Storers have to support regular expression and depending on the implementation of the storage handler the accepted syntax may vary.
+An error of `ErrNotImplemented` will be returned for those storage backends which do not support the `$regex` operator.
+
```
rs: Could you please add a word about how to pass flags here?

omani: done.

PR #76, opened 2017-01-01
jsonschema: Switch to map based encoding

smyrman:
Provides a public interface for #35, and switches the internal encoding
logic to rely on maps, as propsed in #51. This makes the implementation
simpler and more straight forward.

Compared to the poposed solution 1 in #35 that would rely on passing an
io.Writer to an encoder method, this solution will more easily handle a
corner-case where a custom validator writes zero bytes to the io.Writer.

The reduction in complexity does not come for free though. This looks to
be 4x slower for very tiny schemas, and aproximately 2x slower for small
schemas. As jsonschema is used as API documentation, and can be easily
cached, this performance loss should still be quite acceptable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1489,7 +1489,40 @@ fmt.Println(b.String()) // Valid JSON Document describing the schema
  - [ ] schema.Dict
  - [ ] schema.AllOf
  - [ ] schema.AnyOf
- - [ ] Custom validators
+ - [x] Custom FieldValidators
+
+### Custom FieldValidators
+
+For a custom `FieldValidator` to support encoding to JSON Schema, it must implement the `jsonschema.Builder` interface:
+
+```go
+// The Builder interface should be implemented by custom schema.FieldValidator implementations to allow JSON Schema
+// serialization.
+type Builder interface {
+	// BuildJSONSchema should return a map containing JSON Schema Draft 4 properties that can be set based on
+	// FieldValidator data. Application specific properties can be added as well, but should not conflict with any
+	// legal JSON Schema keys.
+	BuildJSONSchema() (map[string]interface{}, error)
+}
+```
+
+To easier extend a `FieldValidator` from the `schema` package, you can call `ValidatorBuilder` inside `BuildJSONSchem()`:
+
+```go
+type Email struct {
+	schema.String
+}
+
+func (e Email) BuildJSONSchema() (map[string]interface{}, error) {
+	parentBuilder, _ = jsonschema.ValidatorBuilder(e.String)
+	m, err := parentBuilder.BuildJSONSchema()
+	if err != nil {
+		return nil, err
+	}
+	m["format"] = "email"
+	return m, nil
+}
+```
 
 ## Licenses
 
--- a/schema/encoding/jsonschema/array.go
+++ b/schema/encoding/jsonschema/array.go
@@ -1,28 +1,29 @@
 package jsonschema
 
-import (
-	"io"
-	"strconv"
+import "github.com/rs/rest-layer/schema"
 
-	"github.com/rs/rest-layer/schema"
-)
+type arrayBuilder schema.Array
 
-func encodeArray(w io.Writer, v *schema.Array) error {
-	ew := errWriter{w: w}
-
-	ew.writeString(`"type": "array"`)
+func (v arrayBuilder) BuildJSONSchema() (map[string]interface{}, error) {
+	m := map[string]interface{}{
+		"type": "array",
+	}
 	if v.MinLen > 0 {
-		ew.writeFormat(`, "minItems": %s`, strconv.FormatInt(int64(v.MinLen), 10))
+		m["minItems"] = v.MinLen
 	}
 	if v.MaxLen > 0 {
-		ew.writeFormat(`, "maxItems": %s`, strconv.FormatInt(int64(v.MaxLen), 10))
+		m["maxItems"] = v.MaxLen
 	}
 	if v.ValuesValidator != nil {
-		ew.writeString(`, "items": {`)
-		if ew.err == nil {
-			ew.err = encodeValidator(w, v.ValuesValidator)
+		builder, err := ValidatorBuilder(v.ValuesValidator)
+		if err != nil {
+			return nil, err
 		}
-		ew.writeString("}")
+		items, err := builder.BuildJSONSchema()
+		if err != nil {
+			return nil, err
+		}
+		m["items"] = items
 	}
-	return ew.err
+	return m, nil
 }
--- a/schema/encoding/jsonschema/encoder.go
+++ b/schema/encoding/jsonschema/encoder.go
@@ -1,7 +1,7 @@
 package jsonschema
 
 import (
-	"fmt"
+	"encoding/json"
 	"io"
 
 	"github.com/rs/rest-layer/schema"
@@ -11,57 +11,37 @@ import (
 // FieldValidator types in the schema package is supported at the moment. Custom validators are also not yet handled.
 // Attempting to encode a schema containing such fields will result in a ErrNotImplemented error.
 type Encoder struct {
-	io.Writer
+	w io.Writer
 }
 
 // NewEncoder returns a new JSONSchema Encoder that writes to w.
 func NewEncoder(w io.Writer) *Encoder {
-	return &Encoder{w}
+	return &Encoder{w: w}
 }
 
 // Encode writes the JSON Schema representation of s to the stream, followed by a newline character.
 func (e *Encoder) Encode(s *schema.Schema) error {
-	ew := errWriter{w: e.Writer}
-	ew.writeString("{")
-	if ew.err == nil {
-		ew.err = encodeSchema(e.Writer, s)
+	m := make(map[string]interface{})
+	if err := addSchemaProperties(m, s); err != nil {
+		return err
 	}
-	ew.writeString("}\n")
-	return ew.err
+	enc := json.NewEncoder(e.w)
+	return enc.Encode(m)
+
 }
 
-// Wrap IO writer so we can consolidate error handling in a single place. Also track properties written so we know when
-// to emit a separator.
-type errWriter struct {
-	w   io.Writer // writer instance
-	err error     // track errors
+// The Builder interface should be implemented by custom schema.FieldValidator implementations to allow JSON Schema
+// serialization.
+type Builder interface {
+	// BuildJSONSchema should return a map containing JSON Schema Draft 4 properties that can be set based on
+	// FieldValidator data. Application specific properties can be added as well, but should not conflict with any
+	// legal JSON Schema keys.
+	BuildJSONSchema() (map[string]interface{}, error)
 }
 
-// Write ensures compatibility with the io.Writer interface.
-func (ew errWriter) Write(p []byte) (int, error) {
-	if ew.err != nil {
-		return 0, ew.err
-	}
-	return ew.w.Write(p)
-}
+// builderFunc is an adapter that allows pure functions to implement the Builder interface.
+type builderFunc func() (map[string]interface{}, error)
 
-func (ew errWriter) writeFormat(format string, a ...interface{}) {
-	if ew.err != nil {
-		return
-	}
-	_, ew.err = fmt.Fprintf(ew.w, format, a...)
-}
-
-func (ew errWriter) writeString(s string) {
-	if ew.err != nil {
-		return
-	}
-	_, ew.err = ew.w.Write([]byte(s))
-}
-
-func (ew errWriter) writeBytes(b []byte) {
-	if ew.err != nil {
-		return
-	}
-	_, ew.err = ew.w.Write(b)
+func (f builderFunc) BuildJSONSchema() (map[string]interface{}, error) {
+	return f()
 }
--- a/schema/encoding/jsonschema/example_test.go
+++ b/schema/encoding/jsonschema/example_test.go
@@ -14,18 +14,20 @@ func ExampleEncoder() {
 	s := schema.Schema{
 		Fields: schema.Fields{
 			"foo": schema.Field{
-				Required: true,
-				// NOTE: Min is currently encoded as '0E+00', not '0'.
+				Required:  true,
 				Validator: &schema.Float{Boundaries: &schema.Boundaries{Min: 0, Max: math.Inf(1)}},
 			},
 			"bar": schema.Field{
 				Validator: &schema.Integer{},
 			},
 			"baz": schema.Field{
-				ReadOnly:  true,
-				Validator: &schema.String{MaxLen: 42},
+				Description: "baz can not be set by the user",
+				ReadOnly:    true,
+				Validator:   &schema.String{MaxLen: 42},
 			},
-			"foobar": schema.Field{},
+			"foobar": schema.Field{
+				Description: "foobar can hold any valid JSON value",
+			},
 		},
 	}
 	b := new(bytes.Buffer)
@@ -35,25 +37,28 @@ func ExampleEncoder() {
 	json.Indent(b2, b.Bytes(), "", "| ")
 	fmt.Println(b2)
 	// Output: {
-	// | "type": "object",
 	// | "additionalProperties": false,
 	// | "properties": {
 	// | | "bar": {
 	// | | | "type": "integer"
 	// | | },
 	// | | "baz": {
+	// | | | "description": "baz can not be set by the user",
+	// | | | "maxLength": 42,
 	// | | | "readOnly": true,
-	// | | | "type": "string",
-	// | | | "maxLength": 42
+	// | | | "type": "string"
 	// | | },
 	// | | "foo": {
-	// | | | "type": "number",
-	// | | | "minimum": 0E+00
+	// | | | "minimum": 0,
+	// | | | "type": "number"
 	// | | },
-	// | | "foobar": {}
+	// | | "foobar": {
+	// | | | "description": "foobar can hold any valid JSON value"
+	// | | }
 	// | },
 	// | "required": [
 	// | | "foo"
-	// | ]
+	// | ],
+	// | "type": "object"
 	// }
 }
--- a/schema/encoding/jsonschema/object.go
+++ b/schema/encoding/jsonschema/object.go
@@ -1,15 +1,28 @@
 package jsonschema
 
 import (
-	"io"
+	"fmt"
 
 	"github.com/rs/rest-layer/schema"
 )
 
-func encodeObject(w io.Writer, v *schema.Object) error {
-	ew := errWriter{w: w}
-	if ew.err == nil && v.Schema != nil {
-		ew.err = encodeSchema(w, v.Schema)
+type objectBuilder schema.Object
+
+var (
+	//ErrNoSchema is returned when trying to JSON Encode a schema.Object with the Schema property set to nil.
+	ErrNoSchema = fmt.Errorf("no schema defined for object")
+)
+
+func (v objectBuilder) BuildJSONSchema() (map[string]interface{}, error) {
+	if v.Schema == nil {
+		return nil, ErrNoSchema
 	}
-	return ew.err
+
+	m := map[string]interface{}{}
+	err := addSchemaProperties(m, v.Schema)
+	if err != nil {
+		return nil, err
+	}
+	return m, nil
+
 }
--- a/schema/encoding/jsonschema/schema.go
+++ b/schema/encoding/jsonschema/schema.go
@@ -1,13 +1,8 @@
 package jsonschema
 
 import (
-	"encoding/json"
 	"errors"
-	"fmt"
-	"io"
 	"sort"
-	"strconv"
-	"strings"
 
 	"github.com/rs/rest-layer/schema"
 )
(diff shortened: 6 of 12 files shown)


smyrman:
@rs, @yanfali. This provides the following public interface for custom validators to implement:

```
+type Formatter interface {
 +	// JSONSchema should return a map containing JSON Schema Draft 4 properties that can be set based on
 +	// FieldValidator data. Application specific properties can be added as well, but should not conflict with any
 +	// legal JSON Schema keys.
 +	JSONSchema() (map[string]interface{}, error)
  }
```

This is, based on "Solution 2" from #35, even you preferred solution 1 there @rs. The internal logic is also switched to rely on maps.

As far as I can see, this solution has the following benefits over solution 1:
- Simpler
- Allows future support for users to encode JSON Schema into other formats than JSON, e.g. YAML as supported by Swagger.
- Avoids a corner case for an io.Writer based solution to #35 where no bytes are written by a custom validator. See [schema.go#L104](https://github.com/smyrman/rest-layer/blob/c671d40ef98b4a37f133f22d8e395377213dd1c7/schema/encoding/jsonschema/schema.go#L104)

It has the following disadvantages compared to solution 1:
- Slower

schema/encoding/jsonschema/schema.go
```diff
@@ -18,132 +13,104 @@ var (
 	ErrNotImplemented = errors.New("not implemented")
 )
 
-// encodeSchema writes JSON Schema keys and values based on s, without the outer curly braces, to w.
-func encodeSchema(w io.Writer, s *schema.Schema) (err error) {
+const (
+	// schemaSizeHint is the maximum number of keys that can be added to a schema map based on schema.Schema.
+	schemaSizeHint = 7
+	// fieldSizeHint is the maximum number of keys that can be added to a schema map based on schema.Field
```
yanfali: I don't think this comment matches the way you use it in the JSONSchema methods. It's actually the minimum and then you add a local value. I think it would be clearer if you updated this comment and used constants within each encoder file so it was clearer what it was you were trying to optimize here.

I'm also not sure how much value we get out of the specifying this hint, as many of the fields we may add are optional. The default allocator size is probably sufficient.

schema/encoding/jsonschema/example_test.go
```diff
@@ -35,25 +34,25 @@ func ExampleEncoder() {
...
-	// | | | "maxLength": 42
+	// | | | "type": "string"
 	// | | },
 	// | | "foo": {
-	// | | | "type": "number",
-	// | | | "minimum": 0E+00
+	// | | | "minimum": 0,
+	// | | | "type": "number"
 	// | | },
 	// | | "foobar": {}
```
yanfali: I don't really like this much. I know it's permitted by the spec but it feels like we should be stricter when we output. At the very least "type": "object".

schema/encoding/jsonschema/schema.go
```diff
@@ -18,132 +13,104 @@ var (
...
-}
+	for key, field := range fields {
+		if field.Required {
+			required = append(required, key)
+		}
+		formatter, err := validatorFormatter(field.Validator)
+		if err != nil {
+			return err
+		}
+		fm, err := formatter.JSONSchema()
```
yanfali: s/fm/fieldMap/
The terseness hurts the readability.

schema/encoding/jsonschema/schema.go
```diff
@@ -18,132 +13,104 @@ var (
...
+		if err != nil {
+			return err
+		}
+		formatField(fm, field)
```
yanfali: this feels a bit backward, adding the meta fields after having created the map but making JSONSchema responsible for allocating the memory is a direct consequence of this. I would suggest renaming formatField. Maybe encodeFields and encodeField? Also I suggest reversing the order of the the parameters; it's often a nicer API that passes in read-only parameters in first and mutable ones last.

schema/encoding/jsonschema/schema.go
```diff
@@ -18,132 +13,104 @@ var (
...
+	required := make([]string, 0, len(fields))
 
-func (fw *fieldWriter) resetPropertiesCount() {
-	fw.propertiesCount = 0
-}
+	for key, field := range fields {
+		if field.Required {
+			required = append(required, key)
+		}
+		formatter, err := validatorFormatter(field.Validator)
```
yanfali: the naming left me a bit confused at first.
```
encoder, err := validatorEncoder(field.Validator)
```

schema/encoding/jsonschema/schema.go
```diff
@@ -18,132 +13,104 @@ var (
...
 	}
-	fw.writeString("}")
-	fw.resetPropertiesCount()
-	return fw.err
 }
 
-// encodeValidator writes JSON Schema keys and values based on v, without the outer curly braces, to w. Note
-// that not all FieldValidator types are supported at the moment.
-func encodeValidator(w io.Writer, v schema.FieldValidator) (err error) {
+func validatorFormatter(v schema.FieldValidator) (f Formatter, err error) {
```
yanfali: early return would probably be slightly clearer. I'm wary of relying on initialized and named returned values for this use case as it may trip up people later.

schema/encoding/jsonschema/schema.go
```diff
@@ -18,132 +13,104 @@ var (
...
 	if v == nil {
-		return nil
+		f = formatterFunc(nilFormatter)
```
yanfali: ```return formatterFunc(nilFormatter), nil```

schema/encoding/jsonschema/schema.go
```diff
@@ -18,132 +13,104 @@ var (
...
+		return
 	}
 	switch t := v.(type) {
+	case Formatter:
+		f = t
```
yanfali: you can leave out this use case if you just return early above.

schema/encoding/jsonschema/schema.go
```diff
@@ -18,132 +13,104 @@ var (
...
+	case *schema.Bool:
+		f = (*boolFormatter)(t)
```
yanfali: return (*boolFormatter)(t), nil

schema/encoding/jsonschema/schema.go
```diff
@@ -18,132 +13,104 @@ var (
...
-		// FIXME: May break the JSON encoding atm. if t.Schema is nil.
-		err = encodeObject(w, t)
-	case *schema.Time:
-		err = encodeTime(w, t)
-	case *schema.Bool:
-		err = encodeBool(w, t)
+		f = (*objectFormatter)(t)
 	default:
-		return ErrNotImplemented
+		err = ErrNotImplemented
```
yanfali: ```return nil, ErrNotImplemented```

schema/encoding/jsonschema/example_test.go
smyrman: it is still doing that (at the end). It's just that the keys will now be ordered alphabetically, so I had to reorder the keys/values in this example.

smyrman: Again, it is still outputing the same, just ordered differently, and a different formating for the numbers (e.g. 0 instead of 0E+00).

yanfali: Sorry comment is about foobar example

schema/encoding/jsonschema/schema.go
smyrman: It was a hard comment to write, which is never a good sign I guess.

It's not the minimum, but the minimum maximum in a way. The minimum number of fields is 0. This number is the maximum number of properties that can be added directly from the `Field` struct, excluding the filed.Validator attribute, which is what I am trying to state...

I will see if there is a better way.

> I'm also not sure how much value we get out of the specifying this hint, as many of the fields we may add are optional. The default allocator size is probably sufficient.

I will try without a size hints, and then rerun the benchmarks:-)

schema/encoding/jsonschema/schema.go
smyrman: I call it formatter in line with the [ResponseForamtter interface](http://rest-layer.io/#custom-response-formatter-sender) from rest-layer. It's not an encoder anymore, as it doesn't encode the data, only format the map which will be encoded later.

yanfali:
Maybe builder then. Formatter suggests final output

schema/encoding/jsonschema/schema.go
smyrman: If we come up with a new name, we should probably change that consistently, e.g. also for formatSchema.

I am not sure encodeFields is better though, as it's no longer encoding data. I will try to think of something...

schema/encoding/jsonschema/schema.go
smyrman: I will change it do use early returns.

schema/encoding/jsonschema/schema.go
smyrman: Not really. The case above is returning a formatter when filed.Validator == nil.

This case is implementing #35 by allowing custom Validators to implement the Formatter interface.

schema/encoding/jsonschema/schema.go
smyrman: Yeah, I will change all of them:-)

schema/encoding/jsonschema/example_test.go
smyrman: ah, foobar. That is what we should encode if the field.Validator is set to nil. We could change the example, but I think it's the correct think to write in this situation. If it's set to nil, we can't enforce it to be an object.

yanfali:
I find setting null in json tends to be worthless. Much better to not set
the field at all and use undefined in JS or OK pattern in go.

I still think for nil it's much better to set the type for consistency at
the receiver when emitting as it's explicit. Implicit behavior tends to get
people doing different things because of interpretation. I'm arguing set
the type to object for the nil case.

smyrman:
> I find setting null in json tends to be worthless. Much better to not set
the field at all and use undefined in JS or OK pattern in go.

True, but that is besides the point...

> I still think for nil it's much better to set the type for consistency at
the receiver when emitting as it's explicit. Implicit behavior tends to get
people doing different things because of interpretation. I'm arguing set
the type to object for the nil case.

Maybe you misunderstand?

I meen if `schema.Field.Validator` is not set or equal to `nil` (in Go), we don't know which _type_ the value may have. The way I read the code in the `schema` package, it may have _any_ type. So it's pretty simple what the `jsonschema` package should do.

The behaviour as it stands is identical to what it was before this change.

The only viable alternative would be to raise an error if we don't want to support Validaor set to nil in the `rest-layer/schema` package, but I personally don't see why we would want that.

smyrman:
On another topic, the last change addresses most of the review comments.

I will do one more commit soon where I remove the size hints. and rerun the benchmarks. I will leave it as one commit for now and rebase later, in case you want to add/runn one more benchmarks @yanfali.

schema/encoding/jsonschema/schema.go
smyrman: renamed to builder (along with all other instances of the word Formatter).

smyrman:
Removing the size hint appears to be a good call @yanfali:

```
benchmark                                                    old ns/op     new ns/op     delta
BenchmarkEncoder/Schema=Student                              20355         20169         -0.91%

benchmark                                                    old allocs     new allocs     delta
BenchmarkEncoder/Schema=Student                              136            136            +0.00%

benchmark                                                    old bytes     new bytes     delta
BenchmarkEncoder/Schema=Student                              7033          6746          -4.08%
sindrero@sjokolade ../rest-layer/schema/encoding/jsonschema (git)-[remotes/smyrman/jsonschema-ma
```

schema/encoding/jsonschema/schema.go
smyrman: Also added a test for this case now btw.

smyrman:
> The only viable alternative would be to raise an error if we don't want to support Validator set to nil in the rest-layer/schema package, but I personally don't see why we would want that.

I will check what the schema package does today. If it does indeed raise an error, we should do the same in the jsonschema package for now at least.

yanfali:
So if a field is defined as nil in schema we are to output a JSONSchema was empty object which is a stand-in for any type? I can sort of understand that interpretation but it doesn't help the consumer of that schema understand that this field member is used for which is effectively ignore this field?

This is the one major weakness of JSON as a specification, lack of comments. If we're going to do that, I would suggest we explicitly put a "description" that describes this behavior for this field. As a consumer of such an API I would be grateful for explicit documentation explaining this rather than leaving me to guess what it means. This wouldn't violate the JSONSchema spec and would be a reasonable compromise. I use JSONSchema as a way to communicate intent to other humans, not just machines. I also feel this better conveys the intent of leaving the Validator nil.

schema/encoding/jsonschema/schema.go
```diff
@@ -18,132 +13,94 @@ var (
...
-	fw.propertiesCount++
-}
+func addFields(m map[string]interface{}, fields schema.Fields) error {
+	props := make(map[string]interface{}, len(fields))
+	required := make([]string, 0, len(fields))
 
-func (fw *fieldWriter) resetPropertiesCount() {
-	fw.propertiesCount = 0
-}
+	for key, field := range fields {
```
yanfali: s/key/fieldName/

schema/encoding/jsonschema/schema.go
```diff
@@ -18,132 +13,94 @@ var (
...
-// written and emits if not the first.
-func (fw *fieldWriter) comma() {
-	if fw.propertiesCount > 0 {
-		fw.writeString(",")
-	}
-	fw.propertiesCount++
-}
+func addFields(m map[string]interface{}, fields schema.Fields) error {
+	props := make(map[string]interface{}, len(fields))
+	required := make([]string, 0, len(fields))
```
yanfali: ```required := []string{}``` should be fine.

schema/encoding/jsonschema/time.go
```diff
@@ -1,13 +1,12 @@
...
-	"github.com/rs/rest-layer/schema"
-)
+type timeBuilder schema.Time
 
-func encodeTime(w io.Writer, v *schema.Time) error {
-	ew := errWriter{w: w}
-	ew.writeString(`"type": "string", "format": "date-time"`)
-	return ew.err
+func (v timeBuilder) JSONSchema() (map[string]interface{}, error) {
+	m := make(map[string]interface{})
```
yanfali: avoid the allocation, package private var and then return it as is.
```
var timeJSONSchema map[string]interface{}
func init() {
    timeJSONSchema = make(map[string]interface{})
    timeJSONSchema["type"] = "string"
    timeJSONSchema["format"] = "data-time"
)
```

Then simply ```return timeJSONSchema, nil```

schema/encoding/jsonschema/array.go
```diff
@@ -1,28 +1,28 @@
...
-)
+type arrayBuilder schema.Array
 
-func encodeArray(w io.Writer, v *schema.Array) error {
-	ew := errWriter{w: w}
-
-	ew.writeString(`"type": "array"`)
+func (v arrayBuilder) JSONSchema() (map[string]interface{}, error) {
+	m := make(map[string]interface{})
+	m["type"] = "array"
```
rs: I would prefer:
```go
m := map[string]interface{}{
    "type": "array",
}
```

schema/encoding/jsonschema/encoder.go
```diff
@@ -11,57 +11,37 @@ import (
...
 }
 
-func (ew errWriter) writeString(s string) {
-	if ew.err != nil {
-		return
-	}
-	_, ew.err = ew.w.Write([]byte(s))
-}
+// builderFunc is an adapter that allows pure functions to implement the Builder interface.
+type builderFunc func() (map[string]interface{}, error)
```
rs: Why don't we expose this?

schema/encoding/jsonschema/object.go
```diff
@@ -1,15 +1,27 @@
...
-		ew.err = encodeSchema(w, v.Schema)
+type objectBuilder schema.Object
+
+var (
+	//ErrNoSchema is returned when trying to JSON Encode a schema.Object with the Schema property set to nil.
+	ErrNoSchema = fmt.Errorf("no schema defined for object")
+)
+
+func (v objectBuilder) JSONSchema() (map[string]interface{}, error) {
+	m := make(map[string]interface{})
```
rs: Same here, prefer `map[string]interface{}{}`. Only use make when you want to pre-allocate.

schema/encoding/jsonschema/array.go
smyrman: sure, just lazy when I removed the sizeHint for the allocation. I will get it done.

schema/encoding/jsonschema/time.go
smyrman: That won't quite work. If new fields are added later, they would be added to the global var. E.g. if one timeJSONSchema field is set to be required, all later instances would as well.

I can do what @rs suggests below instead, and do initialization on the allocation line.

schema/encoding/jsonschema/encoder.go
smyrman: Currently I don't see any way end-users could implement a `jsonschema.Builder` other than letting a `FieldValidator` implementation also implement this interface. I don't see any way in which BuilderFunc could be used by end-users....

schema/encoding/jsonschema/schema.go
smyrman: > it's often a nicer API that passes in read-only parameters in first and mutable ones last.

Not really. Usually it's the mutable one that is passed in first, both in C and in go. See for instance [append](https://golang.org/pkg/builtin/#append), [copy](https://golang.org/pkg/builtin/#copy) or any other example in the standard lib.

smyrman:
> So if a field is defined as nil in schema we are to output a JSONSchema was empty object which is a stand-in for any type? I can sort of understand that interpretation but it doesn't help the consumer of that schema understand that this field member is used for which is effectively ignore this field?

Not necessarily an empty object, just an object without the `"type"` property. Field properties such as "required", "description" and "readOnly" could still appear in the object, if that's what's added to `schema.Field`.

> This is the one major weakness of JSON as a specification, lack of comments. If we're going to do that, I would suggest we explicitly put a "description" that describes this behavior for this field. 

I understand it's confusing, but if we set a default description for some fields and not for others, that becomes an inconsistent and unexpected behaviour. Instead the one specifying the schema should add a fitting description to all of his fields. Hopefully it should be possible to understand that if the "type" property is not set in JSON, it means that any type can be passed in.

smyrman:
> I will do one more commit soon where I remove the size hints. and rerun the benchmarks. I will leave it as one commit for now and rebase later, in case you want to add/run one more benchmark @yanfali.

OK, rebased to one commit as I don't see it's going to be beneficial to do any more benchmarking specifically for the sizeHint removal...

schema/encoding/jsonschema/example_test.go
smyrman: Updated the example to include a new description @yanfali. Hopefully it's at least a bit more clear from the example what's going on now...

schema/encoding/jsonschema/encoder.go
smyrman: PS! I am happy to expose it later, if we can find a real end-user use-case for it.

smyrman:
Ok,  updated the docs as well.

schema/encoding/jsonschema/example_test.go
smyrman: Updated the example. See comment below

schema/encoding/jsonschema/schema.go
yanfali: Quite right, ignore me.

rs:
I like this solution better finaly. GG.

I would rename the method of the `Builder` interface to `BuildJSONSchema`. It is better to recall the name of the interface in a single method interface.

smyrman:
Let's merge #78 first, and I will run the benchmarks.

smyrman:
Benchmark results are  here!

To summarize, we see that:
- A tiny schema (one field) is 5x slower when the target buffer is pre-allocated, does 3x the number of allocations and allocates more than 8x the amount of memory in total.
- A small schema is more than 2x slower, has 2.5x as many allocations, and allocates 5x the amount of memory in total.
- A quite complex schema is just below 2x slower, does less than 2x the amount of allocations, and allocates slightly above 2x the amount of bytes.

Details:
```
benchmark                                                    old ns/op     new ns/op     delta
BenchmarkEncoder/Schema={Fields:{"b":Bool{}}}                1022          5203          +409.10%
BenchmarkEncoder/Schema={Fields:{"s":String{}}}              1005          5168          +414.23%
BenchmarkEncoder/Schema={Fields:{"s":String{MaxLen:42}}}     1285          5831          +353.77%
BenchmarkEncoder/Schema=Simple                               9128          20294         +122.33%
BenchmarkEncoder/Schema=Complex1                             34508         69803         +102.28%
BenchmarkEncoder/Schema=Complex2                             87383         165865        +89.81%

benchmark                                                    old allocs     new allocs     delta
BenchmarkEncoder/Schema={Fields:{"b":Bool{}}}                12             37             +208.33%
BenchmarkEncoder/Schema={Fields:{"s":String{}}}              12             37             +208.33%
BenchmarkEncoder/Schema={Fields:{"s":String{MaxLen:42}}}     14             41             +192.86%
BenchmarkEncoder/Schema=Simple                               55             133            +141.82%
BenchmarkEncoder/Schema=Complex1                             228            430            +88.60%
BenchmarkEncoder/Schema=Complex2                             545            1004           +84.22%

benchmark                                                    old bytes     new bytes     delta
BenchmarkEncoder/Schema={Fields:{"b":Bool{}}}                240           2001          +733.75%
BenchmarkEncoder/Schema={Fields:{"s":String{}}}              224           2001          +793.30%
BenchmarkEncoder/Schema={Fields:{"s":String{MaxLen:42}}}     248           2104          +748.39%
BenchmarkEncoder/Schema=Simple                               1200          6088          +407.33%
BenchmarkEncoder/Schema=Complex1                             6161          19268         +212.74%
BenchmarkEncoder/Schema=Complex2                             14235         44641         +213.60%
```

Command used to produce benchmarks:
```bash
$ go test -run=NONE -bench=. -benchmem -cpu 1
```

Comparison done using [benchmp](https://github.com/golang/tools/tree/master/cmd/benchcmp).

README.md
```diff
@@ -1463,7 +1463,40 @@ fmt.Println(b.String()) // Valid JSON Document describing the schema
...
+// serialization.
+type Builder interface {
+	// BuildJSONSchema should return a map containing JSON Schema Draft 4 properties that can be set based on
+	// FieldValidator data. Application specific properties can be added as well, but should not conflict with any
+	// legal JSON Schema keys.
+	BuildJSONSchema() (map[string]interface{}, error)
+}
+```
+
+To easier extend a `FieldValidator` from the `schema` package, you can call `ValidatorBuilder` inside `BuildJSONSchem()`:
```
smyrman: I have now exposed ValidatorBuilder to be a public function. Let me know if we should do this differently.

smyrman:
@rs, @yanfali: This is ready for merge if you are happy with it.

smyrman:
> @rs, @yanfali: This is ready for merge if you are happy with it.
Hold on -- need to update one docstring after searchreplacing JSONSchema() -> BuildJSONSchema()

smyrman:
OK, ready.

yanfali:
Those numbers are pretty good for the realistic use case. I have no objection to merging.

smyrman:
rebased again due to new commits to master (no changes).

@rs, are you OK with merging this?

Or do you want to consider an alternate implementation?

PR #78, opened 2017-01-02
Add a more complex schema benchmark

yanfali:
- one request and one response schema based off real world schema usage
 - remove allocation of new byte.Buffer from benchmarks, we're not that
 interested in measuring that.

@smyrman here's a more realistic Schema based on something in production.

### DIFF
--- a/schema/encoding/jsonschema/benchmark_complex_test.go
+++ b/schema/encoding/jsonschema/benchmark_complex_test.go
@@ -1,0 +1,202 @@
+package jsonschema_test
+
+import (
+	"github.com/rs/rest-layer/schema"
+	"math"
+	"time"
+)
+
+var (
+	rfc3339NanoDefault = schema.Field{
+		Description: "UTC start time in RFC3339 format with Nano second support, e.g. 2006-01-02T15:04:05.999999999Z",
+		Validator: &schema.Time{
+			TimeLayouts: []string{time.RFC3339Nano},
+		},
+	}
+)
+
+// Default modifies an existing schema.Field with a default value.
+func Default(f schema.Field, v interface{}) schema.Field {
+	f.Default = v
+	return f
+}
+
+// Required modifies an existing schema.Field to be required.
+func Required(s schema.Field) schema.Field {
+	s.Required = true
+	return s
+}
+
+// RFC3339Nano schema.Field template
+func RFC3339Nano() schema.Field {
+	return rfc3339NanoDefault
+}
+
+// String field
+func String(min, max int, description string) schema.Field {
+	return schema.Field{
+		Description: description,
+		Default:     "",
+		Validator: &schema.String{
+			MinLen: min,
+			MaxLen: max,
+		},
+	}
+}
+
+// Integer returns schema.Field template for int validation.
+func Integer(min, max int, description string) schema.Field {
+	return schema.Field{
+		Default:     min,
+		Description: description,
+		Validator: &schema.Integer{
+			Boundaries: &schema.Boundaries{
+				Min: float64(min),
+				Max: float64(max),
+			},
+		},
+	}
+}
+
+// Bool validator helper
+func Bool(description string) schema.Field {
+	return schema.Field{
+		Description: description,
+		Validator:   &schema.Bool{},
+	}
+}
+
+func getComplexSchema1() schema.Schema {
+	mSchema := &schema.Schema{
+		Description: "m",
+		Fields: schema.Fields{
+			"ma": String(0, 64, "ma"),
+			"mb": String(0, 64, "mb"),
+		},
+	}
+	s := schema.Schema{
+		Description: "example request",
+		Fields: schema.Fields{
+			"a": RFC3339Nano(),
+			"b": RFC3339Nano(),
+			"c": String(0, 100, "b"),
+			"d": Default(Integer(0, math.MaxInt64, "c"), 100),
+			"e": Default(Integer(0, math.MaxInt64, "d"), 100),
+			"f": Default(Integer(0, 100, "e"), 100),
+			"g": Default(Integer(0, math.MaxInt64, "f"), 100),
+			"h": Required(String(0, 65535, "h")),
+			"i": Required(String(0, 65535, "i")),
+			"j": Required(String(0, 255, "j")),
+			"k": Required(String(0, 255, "k")),
+			"l": Required(String(0, 65535, "l")),
+			"m": schema.Field{
+				Description: "m",
+				Validator: &schema.Array{
+					ValuesValidator: &schema.Object{
+						Schema: mSchema,
+					},
+				},
+			},
+		},
+	}
+	s.Compile()
+	return s
+}
+
+func getComplexSchema2() schema.Schema {
+	cSchema := &schema.Schema{
+		Description: "c schema",
+		Fields: schema.Fields{
+			"ca": RFC3339Nano(),
+			"cb": String(0, 64, "cb"),
+			"cc": Integer(0, 65535, "cc"),
+			"cd": Integer(0, 65535, "cd"),
+			"ce": Integer(0, math.MaxInt64, "ce"),
+			"cf": Required(String(0, 128, "cf")),
+			"cg": Required(String(0, 128, "cg")),
+			"ch": Integer(0, math.MaxInt64, "ch"),
+			"ci": Integer(0, math.MaxInt64, "ci"),
+			"cj": Integer(0, math.MaxInt64, "cj"),
+			"ck": Integer(0, math.MaxInt64, "ck"),
+		},
+	}
+	gSchema := &schema.Schema{
+		Description: "gSchema",
+		Fields: schema.Fields{
+			"ga": Required(String(0, 64, "ga")),
+			"gb": Required(String(0, 64, "gb")),
+			"gc": Required(String(0, 64, "gc")),
+			"gd": Required(String(0, 64, "gd")),
+			"ge": Required(String(0, 64, "ge")),
+		},
+	}
+	iSchema := &schema.Schema{
+		Description: "iSchema",
+		Fields: schema.Fields{
+			"ia": Integer(0, math.MaxInt64, "ia"),
+			"ib": Integer(0, math.MaxInt64, "ib"),
+			"ic": Integer(0, math.MaxInt64, "ic"),
+			"id": RFC3339Nano(),
+		},
+	}
+	dSchema := &schema.Schema{
+		Description: "dSchema",
+		Fields: schema.Fields{
+			"da": Required(String(0, 128, "da")),
+			"db": Required(String(0, 32, "db")),
+			"dc": RFC3339Nano(),
+			"dd": RFC3339Nano(),
+			"de": Integer(0, 99999, "de"),
+			"df": Integer(0, 100, "df"),
+			"dg": schema.Field{
+				Description: "dg",
+				Validator: &schema.Object{
+					Schema: gSchema,
+				},
+			},
+			"dh": Required(Bool("dh")),
+			"di": schema.Field{
+				Description: "di",
+				Validator: &schema.Object{
+					Schema: iSchema,
+				},
+			},
+		},
+	}
+	s := schema.Schema{
+		Description: "example response",
+		Fields: schema.Fields{
+			"a": Required(String(16, 16, "a")),
+			"b": Required(String(0, 56, "b")),
+			"c": schema.Field{
+				Description: "c",
+				Required:    true,
+				Validator: &schema.Array{
+					ValuesValidator: &schema.Object{
+						Schema: cSchema,
+					},
+				},
+			},
+			"d": schema.Field{
+				Description: "d",
+				Required:    true,
+				Validator: &schema.Array{
+					ValuesValidator: &schema.Object{
+						Schema: dSchema,
+					},
+				},
+			},
+			"e": Required(String(0, 65335, "e")),
+			"f": schema.Field{
+				Validator:   &schema.String{},
+				Description: "f",
+			},
+			"g": schema.Field{
+				Validator:   &schema.String{},
+				Description: "g",
+			},
+		},
+	}
+	s.Compile()
+	return s
+}
--- a/schema/encoding/jsonschema/benchmark_test.go
+++ b/schema/encoding/jsonschema/benchmark_test.go
@@ -38,15 +38,25 @@ func BenchmarkEncoder(b *testing.B) {
 			},
 		},
 		{
-			Name:   `Schema=Student`,
+			Name:   `Schema=Simple`,
 			Schema: studentSchema,
+		},
+		{
+			Name:   `Schema=Complex1`,
+			Schema: getComplexSchema1(),
+		},
+		{
+			Name:   `Schema=Complex2`,
+			Schema: getComplexSchema2(),
 		},
 	}
 	for i := range testCases {
+		buf := bytes.NewBuffer(make([]byte, 2<<19))
 		tc := testCases[i]
 		b.Run(tc.Name, func(b *testing.B) {
 			for i := 0; i < b.N; i++ {
-				enc := jsonschema.NewEncoder(new(bytes.Buffer))
+				buf.Truncate(0)
+				enc := jsonschema.NewEncoder(buf)
 				enc.Encode(&tc.Schema)
 			}
 		})


yanfali:
I've also remove the allocation of byte.Buffer from the benchmarks since that's not we're measuring here and simply truncate the buffer before use. It makes a small but measurable difference.

schema/encoding/jsonschema/benchmark_complex_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"bytes"
+	"github.com/rs/rest-layer/schema"
+	"github.com/rs/rest-layer/schema/encoding/jsonschema"
+	"math"
+	"testing"
+	"time"
+)
+
+// Optional schema field
+func Optional(s schema.Field) schema.Field {
```
smyrman: I would vote for removing this method, and rely only on using `Required()`.

schema/encoding/jsonschema/benchmark_complex_test.go
```diff
@@ -0,0 +1,221 @@
...
+	return f
+}
+
+// Required modifies an existing schema.Field to be required.
+func Required(s schema.Field) schema.Field {
+	s.Required = true
+	return s
+}
+
+func rfc3339Nano() schema.Field {
```
smyrman: This could "safely" be a var, just like [schema.IDField](https://godoc.org/github.com/rs/rest-layer/schema#pkg-variables)

Since it's not a pointer value, it will be copied on assignment, so that you could still do:

```go
...
s := Schema{
        Fields: schema.Fields{
                "x": Required(rfc3339Nano),
        },
}
```

schema/encoding/jsonschema/benchmark_complex_test.go
```diff
@@ -0,0 +1,221 @@
...
+			TimeLayouts: []string{time.RFC3339Nano},
+		},
+	}
+}
+
+// String field
+func String(min, max int, description string) schema.Field {
+	return schema.Field{
+		Description: description,
+		Required:    true,
```
smyrman: Would it not be more appropriate to rely on `Required(String(...))`?

schema/encoding/jsonschema/benchmark_complex_test.go
```diff
@@ -0,0 +1,221 @@
...
+// Bool validator helper
+func Bool(description string) schema.Field {
+	return schema.Field{
+		Description: description,
+		Required:    true,
+		Validator:   &schema.Bool{},
+	}
+}
+
+var (
```
smyrman: All this schemas are left in the namespace... Perhaps it would be better just to leave a single ComplexSchema/complexSchema and initalize it in an `init()` function.

smyrman: Also, would be nice to place it next to the Student schema definition or move the Student schema definition to be in the same place.

schema/encoding/jsonschema/benchmark_complex_test.go
```diff
@@ -0,0 +1,221 @@
...
+	b.Run("request", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			buf.Truncate(0)
+			enc := jsonschema.NewEncoder(buf)
+			enc.Encode(requestSchema)
+		}
+	})
+	b.Run("response", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			buf.Truncate(0)
```
smyrman: Better to use a freshly allocated buffer to get as similar results as possible. Here the first sub-test might grow the buffer, so that when the second test starts, it doesn't need to grow the buffer like the first test did.

smyrman: PS! It's possible to pause the benchmark timer:
```go
b.StopTimer()
buf := new(bytes.Buffer)
b.StartTimer()
```

schema/encoding/jsonschema/benchmark_complex_test.go
```diff
@@ -0,0 +1,221 @@
...
+				Description: "di",
+				Validator: &schema.Object{
+					Schema: iSchema,
+				},
+			},
+		},
+	}
+)
+
+func BenchmarkEncoderComplex(b *testing.B) {
```
smyrman: Why not add these benchmarks as cases to the existing table of sub-tests in `benchmark_test.go`?

schema/encoding/jsonschema/benchmark_test.go
```diff
@@ -42,11 +42,13 @@ func BenchmarkEncoder(b *testing.B) {
 			Schema: studentSchema,
 		},
 	}
+	buf := new(bytes.Buffer)
 	for i := range testCases {
 		tc := testCases[i]
 		b.Run(tc.Name, func(b *testing.B) {
 			for i := 0; i < b.N; i++ {
-				enc := jsonschema.NewEncoder(new(bytes.Buffer))
+				buf.Truncate(0)
```
smyrman: Same here, we still need to allocate a new buffer to avoid benchmarks affecting each other in a bad way.

schema/encoding/jsonschema/benchmark_complex_test.go
smyrman: We could also rename the  Student schema to e.g. "small". Then the benchmark results would read quite well when put in the same table test list.

```
.../Schema=Small
.../Schema=Complex1
.../Schema=Complex2
```

Wheter it's a "student list", "request" or a "response" is really quite irrelevant...

schema/encoding/jsonschema/benchmark_complex_test.go
yanfali: These are general helpers. I actually have a library of validation help functions that have different defaults, I copied these over for expediency. I'm not concerned since they are in the _test framework and not for general use here. I can certainly remove it, it just made it easier to port the code over.

schema/encoding/jsonschema/benchmark_complex_test.go
yanfali: probably, all just ported over from places where default was different.

schema/encoding/jsonschema/benchmark_complex_test.go
yanfali: it's a pretty specific namespace and only run during benchmarking but sure we can clean that up.

schema/encoding/jsonschema/benchmark_complex_test.go
yanfali: I understand but I'm more concerned we are measuring the noise of new, rather than the code under test. One thing specifically I want to see is the effect of all the buffer allocation vs just simply writing to a buffer. The point of the benchmark is to test the code, not new(bytes.Buffer) from the runtime. Turning this into a truncate amortizes the cost once across all the benchmarks and effectively removes it from the test.

schema/encoding/jsonschema/benchmark_complex_test.go
yanfali: I don't really consider the student schema to be complex. This schema approaches the complexity of my actual uses cases in production.

schema/encoding/jsonschema/benchmark_complex_test.go
yanfali: I tried initially but took a while to sort out the import naming and conventions. I can merge them now I have it working.

schema/encoding/jsonschema/benchmark_test.go
yanfali: this is deliberate. The allocations are measurable. Switching to truncate removes them and makes it clearer where the time is being spent.

yanfali:
Here's without the new bytes.Buffer

```
BenchmarkEncoder/Schema={Fields:{"b":Bool{}}}-4         	 1000000	      1027 ns/op	     240 B/op	      12 allocs/op
BenchmarkEncoder/Schema={Fields:{"s":String{}}}-4       	 2000000	       977 ns/op	     224 B/op	      12 allocs/op
BenchmarkEncoder/Schema={Fields:{"s":String{MaxLen:42}}}-4         	 1000000	      1179 ns/op	     248 B/op	      14 allocs/op
BenchmarkEncoder/Schema=Simple-4                                   	  200000	      8489 ns/op	    1200 B/op	      55 allocs/op
BenchmarkEncoder/Schema=Complex1-4                                 	   50000	     31846 ns/op	    6162 B/op	     228 allocs/op
BenchmarkEncoder/Schema=Complex2-4                                 	   20000	     79457 ns/op	   14237 B/op	     545 allocs/op
```

Here's with

```
BenchmarkEncoder/Schema={Fields:{"b":Bool{}}}-4         	 1000000	      1243 ns/op	     496 B/op	      14 allocs/op
BenchmarkEncoder/Schema={Fields:{"s":String{}}}-4       	 1000000	      1243 ns/op	     480 B/op	      14 allocs/op
BenchmarkEncoder/Schema={Fields:{"s":String{MaxLen:42}}}-4         	 1000000	      1639 ns/op	     504 B/op	      16 allocs/op
BenchmarkEncoder/Schema=Simple-4                                   	  200000	      9659 ns/op	    2528 B/op	      59 allocs/op
BenchmarkEncoder/Schema=Complex1-4                                 	   50000	     34222 ns/op	    9827 B/op	     233 allocs/op
BenchmarkEncoder/Schema=Complex2-4                                 	   20000	     84637 ns/op	   26321 B/op	     552 allocs/op
```

schema/encoding/jsonschema/benchmark_test.go
```diff
@@ -51,7 +51,8 @@ func BenchmarkEncoder(b *testing.B) {
 			Schema: responseSchema,
 		},
 	}
-	buf := new(bytes.Buffer)
+	buf := &bytes.Buffer{}
+	buf.Grow(65535)
```
rs: Wouldn't it be cleaner to write `buf := bytes.NewBuffer(make([]byte, 65535))`?

schema/encoding/jsonschema/benchmark_complex_test.go
smyrman: Even test code should be clean, perhaps even more clean than the code itself. Tests also serve as documentation you know.

I understand that you copy-paste, but it would be cleaner to leave required = false as the default and remove the Optional method.

schema/encoding/jsonschema/benchmark_complex_test.go
smyrman: > it's a pretty specific namespace and only run during benchmarking but sure we can clean that up.

The namespace is the same for the entire jsonschema_test package. Anyway thanks for cleaning it up.

> I don't really consider the student schema to be complex.

No, which is why I suggest you call it "Small":-)

schema/encoding/jsonschema/benchmark_test.go
smyrman: See the docs:

> Truncate discards all but the first n unread bytes from the buffer but continues to use the same allocated storage.

So at least we need to allocate a new buffer per sub-test, even if we don't do so within the loop itself...  That should eliminate the chance of Grow affecting different sub-tests differently.

Or use `ioutil.Discard` as the writer.

schema/encoding/jsonschema/benchmark_complex_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+// Bool validator helper
+func Bool(description string) schema.Field {
+	return schema.Field{
+		Description: description,
+		Validator:   &schema.Bool{},
+	}
+}
+
+func getComplexSchema() (schema.Schema, schema.Schema) {
```
smyrman: It's not obvious why this method need to return two schemas, and why we want to call them requestSchema and responseSchema.

I would suggest moving everything to `benchmark_test.go` file, and initialize it like this:

```go
// reusable Schemas for benchmarks and tests.
var (
        schemaSmall schema.Schema
        schemaComplex1 schema.Schema
        schemaComplex2 schema.Schema
)

func init() {
        // Initalize the schemas:
        // schemaSmall = todays Student schema (or one with similar complexity).
        // schemaComplex1 = requestSchema
        // schemaComplex2 = responseSchema
}
```

After that, replace any reference to Student with Small in the other tests.

schema/encoding/jsonschema/benchmark_test.go
```diff
@@ -38,15 +39,25 @@ func BenchmarkEncoder(b *testing.B) {
...
+		{
+			Name:   `Schema=Complex1`,
+			Schema: requestSchema,
+		},
+		{
+			Name:   `Schema=Complex2`,
+			Schema: responseSchema,
+		},
 	}
+	buf := bytes.NewBuffer(make([]byte, 65535))
```
smyrman: This helps, but if any one benchmark was to trigger a Grow (which usually means doubling the buffer size I think), that one benchmark would be unfairly punished. Maybe it's better to just rely on `ioutil.Discard`? Then we don't need to worry about the time spent allocating/growing the buffer at all. Of course, the write itself also becomes free.

Alternatively, we should at the very least move this to inside `for i := range testCases` (before the b.Run call), to assure each benchmarks starts with a writer that has the same capacity.

smyrman: I think relying on [ioutil.Discard](https://golang.org/pkg/io/ioutil/#pkg-variables) is probably the best option btw.

schema/encoding/jsonschema/benchmark_complex_test.go
smyrman: > I would suggest moving everything to benchmark_test.go file

On second thought.. maybe just rename the file to `testutils_test.go`. That would be a good place for this helper functions, and perhaps also some helper functions from `all_test.go` should move here later.

yanfali: I disagree with your suggestion, now you're back to polluting the name space. You can't have it both ways. Sure we can rename the file but then you loose the intent of what the file was for.

yanfali: getComplexSchema is a template generator for two schemas based off actual production code. The fact the signature returns multiple schema is irrelevant; it was done for name spacing reasons. I have an idea that may covey intent better, if we return the test structs directly maybe?

schema/encoding/jsonschema/benchmark_test.go
yanfali: Sorry, but why would anything call grow? we control the buffer and we only pass it in as an io.Writer. I like the Discard idea, but it could make debugging harder if we needed to pin point any problems. Discard would also not trigger any of the io.Writer error behavior since it's essentially a null instance. Truncate has the behavior I'm after. It's resets the buffer without doing any new allocations.

```
    67	func (b *Buffer) Truncate(n int) {
    68		b.lastRead = opInvalid
    69		switch {
    70		case n < 0 || n > b.Len():
    71			panic("bytes.Buffer: truncation out of range")
    72		case n == 0:
    73			// Reuse buffer space.
    74			b.off = 0
    75		}
    76		b.buf = b.buf[0 : b.off+n]
    77	}
```

schema/encoding/jsonschema/benchmark_test.go
yanfali: Sorry but this is a strawman argument. we completely control the test loop. The encoder is only using the io.Writer interface.

schema/encoding/jsonschema/benchmark_test.go
yanfali: On reflection, I actually think the Discard idea is probably not a good idea because the writes are a part of the performance of the API. The memory bandwidth of the test host would affect the outcome and if we use Discard we wouldn't get a measurement of that. The only thing these changes do is hoist of the allocation Buffer itself outside of the measurement loop.

schema/encoding/jsonschema/benchmark_complex_test.go
smyrman: I where not convened with the so called "requestSchema" and "responseSchema" polluting the namespace but about all the intermediate ones.

The suggested names complexSchema1/2 could be useful for reuse in tests.

> getComplexSchema

Even test code should be clean. Perhaps even cleaner than the main code, since tests also serve as examples/documentation to package maintainers and users.

I believe a good clean function should do one and only one thing. I think this mwthod does two which makes it unclean, and it makes it harder than it has to be to reuse the complex schema for tests

schema/encoding/jsonschema/benchmark_test.go
smyrman: > Sorry, but why would anything call grow? 

If any schema ends up bigger than the prealocated buffer, Grow will be called implisetly.

smyrman: To quote the docs:

> Write appends the contents of p to the buffer, growing the buffer as needed.

schema/encoding/jsonschema/benchmark_complex_test.go
yanfali: Good argument. I will separate this in to two helper functions.

As an aside, this is test code. Getting too attached to excessively "clean" test code is among the worst mistakes I feel you can make because it it can become a burden on the developers; with test code you must always be prepared to completely throw it out, it's not your product - it's an expression of the current state of the implementation that helps you have comfort that the code is doing what you intended.

I do agree it should serve as a form of documentation though, but by conveying the intent.

schema/encoding/jsonschema/benchmark_test.go
yanfali: yes I see your argument, but we also control the test environment. At worst we can increase the initial buffer size to 1MiB. I don't see any realistic schema ever growing beyond that size.

yanfali: This is interesting. Setting the buffer to 2<<20 made the test worse. Setting the buffer to 2<<19 actually made it improve. It's probably some hardware related caching, but it's still interesting. I'm on a 13" macbook pro mid 2014.

smyrman: The _best option_ would actually be to use a buffer that panicked instead of automatically growing, so that we were warned if a growing of buffers were triggered. I don't know if any such buffer exists in the standard lib?

If that's not possible, we can at the very least move it within the first loop like this:

```go
for i := range testCases {
	tc := testCases[i]
	buf := bytes.NewBuffer(make([]byte, 65535))
 	b.Run(tc.Name, func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			buf.Truncate(0)
			...
```

This _ensures_ that benchmarks will not interfere with each other.

smyrman: Actually, we could also just run the encoding once before the sub-benchmark loop starts! That would also ensure we always have enough space!

yanfali: I'm happy to move the buffer inside the test harness loop and outside of the benchmark loop. That's a reasonable compromise. I don't know if we want to be too clever in a benchmark or test code. I find it just gets you into trouble :)

smyrman:
```go
for i := range testCases {
	tc := testCases[i]
	enc := jsonschema.NewEncoder(buf)
	// We don't want buf grow to affect the benchmark, so we run through the encoding
	// once before the benchmark starts.
	buf.Truncate(0)
	enc.Encode(&tc.Schema)
 	b.Run(tc.Name, func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			buf.Truncate(0)
			enc.Encode(&tc.Schema)
			...
```

schema/encoding/jsonschema/benchmark_test.go
smyrman: (See discussion above instead)

schema/encoding/jsonschema/benchmark_complex_test.go
smyrman: > I will separate this into two helper functions.

Better👍

yanfali:
Performance appears the same, we're just making the GC work a bit harder.

smyrman:
> Performance appears the same, we're just making the GC work a bit harder.

Did you see my last suggestion?

We could run the Encode once before the b.Run starts() (code example in inline comment). Then we only allocate just-enough bytes, and we reuse the buffer, but we always do it before the benchmarks start so it's not counted.

yanfali:
I did, though I'm not sure it's worth the extra complexity, let me run a quick benchmark and see how it compares. If it's roughly the same I think we should just leave it out.

yanfali:
Over three runs it's about the same performance as before, so I'd vote to keep it as straightforward as possible and this avoids having the extra comment explaining why we are doing this.

smyrman:
> Over three runs it's about the same performance as before, so I'd vote to keep it as straightforward as possible and this avoids having the extra comment explaining why we are doing this.

Sounds good. And thanks a lot for the  new benchmarks. I will re-test #76 once this is merged :-)

yanfali:
@rs this is ready to go, I'm going to squash it first.

schema/encoding/jsonschema/benchmark_complex_test.go
```diff
@@ -0,0 +1,203 @@
...
+				},
+			},
+		},
+	}
+	s.Compile()
+	return s
+}
+
+func getComplexSchema2() schema.Schema {
+
```
rs: Extra return

yanfali: sorry? a bit confused by this comment.

rs: There is an extra blank line.

yanfali:
@rs thanks fixed. It was close enough to the `return` statement above that I was confused by the comment.

Issue #33, opened 2016-08-21
[jsonschema] Support for schema.Dict


smyrman:
As suggested by @yanfali, I am adding tickets for how to extend the initial JSON Schema support to support more types. `schema.Dict` will get it's own ticket, as it's tricky enough to deserve so.
# Forward

`schema.Dict` should be encoded to the same JSON type as `schema.Object` (object), but rely on `additionalProperties` when a  `KeysValidator` is not set or `patternProperties` if `KeyValidator` _is_ set.

According to the validaiton spec [section 5.4.4](http://json-schema.org/latest/json-schema-validation.html#anchor64), `aditionalProperties` can either be a boolen, in which case any value is allowed when it's `true`, or a JSON Schema, in which case the values must apply to that schema. `patternProperties`  must be an object with regex as keys, and JSON Schemas as values.

However, as far as I can see from the examples, as well as from the [Core/Validation Meta-Schema](http://json-schema.org/schema), there are no required elements in a JSON Schema object, which means it can be left emty in order to allow for all types (similar to the empty `interface{}` in go ).
# Expected result

This really cumulates to us wanting to end up with the following.

If neither`KeyValidator` nor `ValueValidator` is set:

``` json
"type": "object",
"additionalProperties": true
```

If only `ValueValidator` is set:

``` json
"type": "object",
"additionalProperties": {/* JSON Schema from validatorToJSONSchema(ValuesValidator) */}
```

If only `KeyValidator` is set:

``` json
"type": "object",
"patternProperties": {
    "<regex from KeyValidator>": {}
}
```

If both `KeyValidator` and `ValueValidator` is set

``` json
"type": "object",
"patternProperties": {
    "<regex from KeyValidator>": {/* JSON Schema from validatorToJSONSchema(ValuesValidator) */}
}
```

rs:
Isn't this already fixed with `schema.Object`?

smyrman:
No, schema.Dict is not yet handled. The most tricky part will be to handle the key validatior, as it's likely that we can only handle it properly if it's the builtin string type and uses regex (or something we canvert to regex).

There is plans in JSON Schema to support something more than just "patternProperties", but that's not ready yet:
https://github.com/json-schema-org/json-schema-spec/issues/117

smyrman:
I will start looking at an implementation for this issue.

PR #84, opened 2017-01-22
Issue #34: JSON Schema support for more types

### LINKED ISSUE #34: [jsonschema] Support for additional types
This ticket aims to add support for translating unhandled validators from the `schema` package to a JSON Schema type (with an appropriate "format"). It is assumed that the JSON fields below are added to the same JSON object that is created for the parent `schema.Field` type.
# Relevant background on "format"

"format" is covered by the validation spec [section 7.  Semantic validation with "format"](http://json-schema.org/latest/json-schema-validation.html#anchor104). This spec specifies some formats, as well as allowing implementations to define custom formats when needed. Where possible, I will fall back to suggest a format from [The Swagger 2.0 spec](http://swagger.io/specification/) if the JSON spec fails to define any format in particular, and the format would provide useful information for end-users or tools.
# String types

For `schema.IP`, JSON Schema defines only "IPv4" and "IPv6". The most correct here is to leave the type in the outer schema, and then  use a "oneOf" clause to encapsulate the "format" part of the schema as a "sub-schema":

``` json
"type": "string",
"oneOf": [
    {"format": "IPv4"},
    {"format": "IPv6"},
]
```

For `schema.URL` we could go for just:

``` json
"type": "string",
"format": "uri"
```

This ignores all the validation options that might be set in a `schema.URL` package, which is not ideal, but perhaps good enough. While this is a pragmatic solution, the correctness of this implemenetation might be argued, as the spec staes everythin accepted as a valid URI by https://tools.ietf.org/html/rfc3986.

For `schema.Password` we could go for:

``` json
"type": "string",
"format": "password"
```

Just like for `schema.String`, "minLength" and "maxLength" should be encoded from `minLen` and `maxLen` when set.

Note that the "password" format is not defined by JSON Schema. It is however [defined by Swagger 2.0](http://swagger.io/specification/#dataTypeFormat), and it seams reasonable to align with Swagger on extensions when possible.

The purpose of the format according to Swagger 2.0: "Used to hint UIs the input needs to be obscured."
# Other validators

`schema.Null` should simply be represented as:

``` json
"type": "null"
```

This type is documented in the core spec [section 3.5. JSON Schema primitive types](http://json-schema.org/latest/json-schema-core.html#anchor8)

smyrman:
Fixes #34 by adding JSON Schema support for FieldValidators:
- Null
- Password
- IP
- URL (limited support)

As for URL, none of the validation parameters in schema.URL is taken
into account by the JSON Schema encoding at this point in time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1476,14 +1476,14 @@ fmt.Println(b.String()) // Valid JSON Document describing the schema
 
  - [x] schema.Schema
  - [x] schema.Bool
- - [ ] schema.Null
+ - [x] schema.Null
  - [x] schema.Float
  - [x] schema.Integer
  - [x] schema.String
  - [x] schema.Time
- - [ ] schema.URL
- - [ ] schema.IP
- - [ ] schema.Password
+ - [x] schema.URL (limited support)
+ - [x] schema.IP
+ - [x] schema.Password
  - [x] schema.Array
  - [x] schema.Object
  - [ ] schema.Dict
--- a/schema/encoding/jsonschema/ip.go
+++ b/schema/encoding/jsonschema/ip.go
@@ -1,0 +1,16 @@
+package jsonschema
+
+import "github.com/rs/rest-layer/schema"
+
+type ipBuilder schema.IP
+
+func (v ipBuilder) BuildJSONSchema() (map[string]interface{}, error) {
+	m := map[string]interface{}{
+		"type": "string",
+		"oneOf": []map[string]interface{}{
+			{"format": "ipv4"},
+			{"format": "ipv6"},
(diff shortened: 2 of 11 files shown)


schema/encoding/jsonschema/bool_test.go
```diff
@@ -1,5 +1,3 @@
-// +build go1.7
-
```
smyrman: btw, this removes some build go1.7 tags.

smyrman:
@rs - should be ready to merge.

When this is merged I will push a small cleanup PR before staring on #33.

PR #88, opened 2017-02-01
JSON Schema support for schema.Dict

### LINKED ISSUE #33: [jsonschema] Support for schema.Dict

smyrman:
This commits resolves #33 by adding JSON Schema support for schema.Dict.
The support limits KeysValidator instances to be a schema.String
instance or nil as explained in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1459,7 +1459,7 @@ See [Hystrix godoc](https://godoc.org/github.com/afex/hystrix-go/hystrix) for mo
 
 ## JSONSchema
 
-An incomplete but stil useful JSONSchema implementation that covers many common use cases for Schema. Goal is to try and match the draft-04 spec. Patches welcome.
+An incomplete but still useful JSONSchema implementation that covers many common use cases for Schema. Goal is to try and match the draft-04 spec. Patches are welcome.
 
 ```go
 import "github.com/rs/rest-layer/schema/jsonschema"
@@ -1474,11 +1474,12 @@ fmt.Println(b.String()) // Valid JSON Document describing the schema
 
 ### Supported FieldValidators
 
+ - [x] Custom FieldValidators
  - [ ] schema.AllOf
  - [ ] schema.AnyOf
  - [x] schema.Array
  - [x] schema.Bool
- - [ ] schema.Dict
+ - [x] schema.Dict (limited support)
  - [x] schema.Float
  - [x] schema.IP
  - [x] schema.Integer
@@ -1489,7 +1490,6 @@ fmt.Println(b.String()) // Valid JSON Document describing the schema
  - [x] schema.String
  - [x] schema.Time
  - [x] schema.URL (limited support)
- - [x] Custom FieldValidators
 
 ### Custom FieldValidators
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1489,40 +1490,51 @@ fmt.Println(b.String()) // Valid JSON Document describing the schema
  - [x] schema.String
  - [x] schema.Time
  - [x] schema.URL (limited support)
- - [x] Custom FieldValidators
 
 ### Custom FieldValidators
 
-For a custom `FieldValidator` to support encoding to JSON Schema, it must implement the `jsonschema.Builder` interface:
+ For a custom `FieldValidator` to support encoding to JSON Schema, it must implement the `jsonschema.Builder` interface:
```
smyrman: Opps extra space...

schema/encoding/jsonschema/dict_test.go
```diff
@@ -0,0 +1,290 @@
...
+			customValidate: fieldValidator("d", `{
+				"type": "object",
+				"additionalProperties": false,
+				"patternProperties": {
+					"re": {}
+				}
+			}`),
+		},
+		{
+			name: `KeyValidator=String{Regex:"re"},ValuesValidator=Integer{}"`,
```
smyrman: should be Regexp in name

schema/encoding/jsonschema/dict_test.go
```diff
@@ -0,0 +1,290 @@
...
+		},
+		{
+			name: `KeyValidator=String{Allowed:["match1"]}"`,
+			schema: schema.Schema{
+				Fields: schema.Fields{
+					"d": {
+						Validator: &schema.Dict{
+							KeysValidator: &schema.String{
+								Regexp:  "tch",
+								Allowed: []string{"match1", "match2"},
```
smyrman: This test should not have passed...

smyrman:
Let's wait for #90 to be merged first and then rebase.

PR #91, opened 2017-02-01
Minor README fix

smyrman:
uriref link was formated wrong.


smyrman:
Seems I formated this wrong in #88. Sorry about that.

Issue #57, opened 2016-12-10
Validate schema.Reference when embedded in nested FieldValidators


omani:
Hi,

I have the following schema:
```go
var Roles = schema.Schema{
	Fields: schema.Fields{
		"id":      schema.IDField,
		"created": schema.CreatedField,
		"updated": schema.UpdatedField,

		// fields
		"name": {
			Required:   true,
			Filterable: true,
			Sortable:   true,
			Validator: &schema.String{
				MinLen: 1,
				MaxLen: 63,
			},
		},
		"description": {
			Filterable: true,
			Sortable:   true,
			Validator: &schema.String{
				MaxLen: 63,
			},
		},
		"usernames": {
			Validator: &schema.Reference{
				Path: "users",
			},
		}
}
```

I would like to have usernames as an array of references. but I want to be able to pass the user's name instead of its IDField. something like:

```
pseudo POST request:
...
"usernames": ["omani", "rs"]
...
```

how do I do this? I guess there is no support for this in the Validator and I have to do this in a hook?

omani:
maybe it is easier to use the `users` schema as a subresource to `usernames`. does this make more sense?

afaiu this results in having nested documents in mongodb instead of references, right?

omani:
I've implemented a MutlGetByField method to be able to fetch for a subset of paths.

If anyone is interested you will find it [here](https://github.com/omani/rest-layer/commit/822fdef3ff40c41aaa72cdb0b632c599a3c27afa)

I still have to figure out how to have arrays of references though.

omani:
Also see my latest commit [here](https://github.com/omani/rest-layer/commit/db42cb0fd39864cfe958fbec5a4adc31d919c280) which provides a new schema calles `ReferenceArray`.

smyrman:
> Also see my latest commit here which provides a new schema calles ReferenceArray.

As for the `ReferenceArray` type, would not a normal `Array` validator with `ValuesValidator` set to `Reference` do the trick?
```go
...
"usernames": {
            Validator: &schema.Array{
                    ValuesValidator: &schema.Reference{
                            Path: "users",
                    },
            },
},
...
```
Or am I missing something?

omani:
I thought about that also. but with that example the referencing does not work. That's why I came up with the ReferenceArray Validator.

smyrman:
> I thought about that also. but with that example the referencing does not work
Sounds like a bug to me... @rs, would you agree?

[...]

PR #100, opened 2017-03-20
Allow schema.Reference to validate as normal field

smyrman:
Solves issue #57 by allowing schema.Reference to be validated as a normal field. See commit messages for more details!


schema/reference.go
```diff
@@ -1,12 +1,23 @@
...
 
-// Validate validates and normalize reference based value
+// Validate validates and sanitizes IDs against the reference path.
 func (v Reference) Validate(value interface{}) (interface{}, error) {
-	// All the work is performed in rest.checkReferences()
-	return value, nil
+	return v.rl.ValidateID(v.Path, value)
+}
+
+// CompileReference validates v.Path against rl and stores rl for later use by v.Validate.
```
smyrman: Should be just Compile  in this implementation example. This is leftover code

schema/reference.go
```diff
@@ -1,12 +1,23 @@
...
+func (v Reference) CompileReference(rl ResourceLookup) error {
```
smyrman: Method should be named Compile(rl ResourceLookup)

smyrman:
One of my big question with this design idea, is weather there will be more objects that various FielValidator's Compile methods needs access to in the future... If no, then this design might be OK.

smyrman:
TL;DR: I will complete this design to facilitate discussion. It's either _this_ or passing context.Context to the Validate method (where the context would contain the same wrapped resource.Index that in effect implements ResourceLookup).

Design discussion: I had been hoping from some comments by now, but I suppose my code was not complete enough to facilitate a full understanding of the intended design here; especially as I *have not implemented* the part that must _wrap_ resource.Index into an implementation of the _ResourceLookup_ interface before passing that in to the `Schema.Compile()` method. Anyway, I have done some thinking myself below that I will outline below.

I have one thing that talks in favor of this desing atm., and that is that it makes sense to validate that the linked Reference resource exist at compile time. It can even help users avoid some bugs from reaching production.

I mentioned on #57 that I have _considered_ replacing the `ResourceLookup` passed to the `Compiler` method with a `context.Context` that can _optionally_ contain a `ResourceLookup`. That however feels _really iffy_, as the main purpose of context.Context is _not_ to pass arbitrary data to (deeply nested) functions, but to facilitate deadlines and cancelation. And that is something the Compiler functions should not need!

The _Validate_ function however, could serve well as a method where `context.Context` could be passed in -- as it could actually make _sense_ to allow for deadline management there -- especially if having to do a DB call to finish the validation, as would be the case for `Reference`.

I will complete the current design to facilitate discussion, but it's probably a viable alternative to consider going for a `Valdiate(ctx context.Context)` interface.

smyrman:
I am not to far away from pushing and update... almost got there in the weekend, just need to update some remaining tests...

rs:
Sorry if I'm not very responsive. I'll wrap my head around the subject when you send the PR.

smyrman:
No problem @rs, I have been very busy as well, hench the long delay in pushing any more code.

smyrman:
@rs, There is still some tests that needs updating, but now at least the tests within the schema package pass.

At least tests within the schema/encoding and resource  packages needs updating.

resource/resource.go
```diff
@@ -111,16 +111,16 @@ func (r *Resource) ParentField() string {
 	return r.parentField
 }
 
-// Compile the resource graph and report any error
-func (r *Resource) Compile() error {
+// Compile the resource graph and report any error.
+func (r *Resource) Compile(rl schema.ReferenceChecker) error {
```
smyrman: sholud be rc for consistency. the rl name is leftover from when it was called ReferenceLookup

smyrman:
UPDATE: jsonschema tests fixed, although I will add better support for JSON Schema encoding of Reference later now that we can actually access the reference after Compile is called.

rest package tests are still failing though, with the following:

```
2017/05/08 07:59:08 Server error: No Storage Defined
--- FAIL: TestHandlerPostListWithReferenceNoRouter (0.00s)
        Error Trace:    method_post_test.go:229
	Error:  	Not equal: 500 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:232
	Error:  	Not equal: 500 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:233
	Error:  	Not equal: "Router not available in context" (expected)
			        != "Document contains error(s)" (actual)
		
--- FAIL: TestHandlerPostListWithInvalidReference (0.00s)
        Error Trace:    method_post_test.go:255
	Error:  	Not equal: 500 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:258
	Error:  	Not equal: 500 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:259
	Error:  	Not equal: "Invalid resource reference for field `foo': invalid" (expected)
			        != "Document contains error(s)" (actual)
		
.... more failures like this.
```

I am not quite sure why, but presumably it's somehow related to issues with index.Compile() ?

resource/index.go
```diff
@@ -5,22 +5,22 @@ import (
 	"net/url"
 	"strings"
 
+	"golang.org/x/net/context"
```
rs: wrong `context`?

schema/reference.go
```diff
@@ -1,12 +1,36 @@
...
+// Compile validates v.Path against rc and stores the a FieldValidator for later use by v.Validate.
+func (r *Reference) Compile(rc ReferenceChecker) error {
+	if rc == nil {
+		return fmt.Errorf("rc can not be nil")
+	}
+
+	if v := rc.ReferenceChecker(r.Path); v != nil {
+		r.validator = v
+		return nil
+
```
rs: extra empty line

resource/index.go
smyrman: yes

smyrman:
@rs, thanks for the review, and sorry for my slow update time on this. Pushing a minor update to resolve identified issues, but there is still one major bug in the resource package that needs some debugging. Reproduce by running tests.

If someone with some time on their hans want to try and debug it, feel welcome. Otherwise I will get to it eventually.

resource/resource.go
smyrman: fixed

resource/index.go
smyrman: fixed

schema/reference.go
smyrman: fixed

smyrman:
Sorry for the the delay in looking into this.

I think most of the errors looks pretty reasonable, since the code generating the error texts / status codes we test against here, is part of what has been removed.

smyrman:
Right, so the 201 issues are solved by adding an explicit Compile to `index`. Normaly, index is compiled by `rest.NewHandler(index)`, but that method is not called in these tests. Previously, this tests worked because a schema compilation was not specifically needed for `schema.Reference` fields, or any of the other fields in this specific schemas.

This leaves the remaining ssues with unexpected status codes / error messages when POSTing documents due to changed semantics. These errors also lacks an explicit call to index.Compile(). When correcting the tests, I think we can wrote them to be go 1.7 sub-tests, so that there is less code duplication.

@rs, what's your opinion on changing the semantics for "invalid reference" to all return a 422 (Unprocessable Entity), rather than different type of errors based on _what's_ wrong with the Reference field. The reason that the status code is changing, is that invalid Reference fields now follow the same code pattern as other validation issues.

```
--- FAIL: TestHandlerPostListWithReferenceNoRouter (0.00s)
        Error Trace:    method_post_test.go:229
	Error:  	Not equal: 500 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:232
	Error:  	Not equal: 500 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:233
	Error:  	Not equal: "Router not available in context" (expected)
			        != "Document contains error(s)" (actual)
		
--- FAIL: TestHandlerPostListWithInvalidReference (0.00s)
        Error Trace:    method_post_test.go:255
	Error:  	Not equal: 500 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:258
	Error:  	Not equal: 500 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:259
	Error:  	Not equal: "Invalid resource reference for field `foo': invalid" (expected)
			        != "Document contains error(s)" (actual)
		
--- FAIL: TestHandlerPostListWithReferenceOtherError (0.00s)
        Error Trace:    method_post_test.go:282
	Error:  	Not equal: 500 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:285
	Error:  	Not equal: 500 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:286
	Error:  	Not equal: "Error fetching resource reference for field `foo': No Storage Defined" (expected)
			        != "Document contains error(s)" (actual)
		
--- FAIL: TestHandlerPostListWithReferenceNotFound (0.00s)
        Error Trace:    method_post_test.go:309
	Error:  	Not equal: 404 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:312
	Error:  	Not equal: 404 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:313
	Error:  	Not equal: "Resource reference not found for field `foo'" (expected)
			        != "Document contains error(s)" (actual)
		
--- FAIL: TestHandlerPostListWithSubSchemaReferenceNotFound (0.00s)
        Error Trace:    method_post_test.go:407
	Error:  	Not equal: 404 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:410
	Error:  	Not equal: 404 (expected)
			        != 422 (actual)
		
        Error Trace:    method_post_test.go:411
	Error:  	Not equal: "Resource reference not found for field `foo'" (expected)
			        != "Document contains error(s)" (actual)
```

rs:
Yes I think it make sense. And for Go 1.7 sub-test, I'm all for it.

smyrman:
I have set the sub-tests to run in Parallel. This appears to uncover a data race within `github.com/rs/rest-layer/rest.DefaultResponseFormatter.FormatError()`, but I don't _think_ it's related to this change.

The bug goes away if I remove the parallel execution of the sub-tests. I think the bug was just masked before because we didn't do to many parallel tests...

smyrman:
It's these lines:
```go
func (f DefaultResponseFormatter) FormatError(ctx context.Context, headers http.Header, err error, skipBody bool) (context.Context, interface{}) {
        oldLogger := resource.Logger
        resource.Logger = nil
        defer func() { resource.Logger = oldLogger }()
```

It think this is pretty ugly:-/ If we really need to do this, we need to use a mutex. What's the benfit of silencing the logger? Do you want to fix it in a separate commit?

This PR should be otherwise good to go in now :-)

rs:
I did that in some tests to remove error output while testing error cases? Did I do that in actual code??

smyrman:
> Did I do that in actual code??
Well, I didn't run a git blame on it yet, but someone did... if it's by mistake I can add a separate commit to remove it in this PR so that the tests pass.

And we can of course silence the loggers in test code (or re-route it to write to testung.T).

smyrman:
The Git-blame btw:
- https://github.com/rs/rest-layer/commit/4d8102671d3b82ac56d174b1ab1d0600cca0662f

schema/anyof.go
```diff
@@ -25,8 +25,7 @@ func (v AnyOf) Compile(rc ReferenceChecker) error {
 // Validate ensures that at least one sub-validator validates.
 func (v AnyOf) Validate(value interface{}) (interface{}, error) {
 	for _, validator := range v {
-		var err error
-		if value, err = validator.Validate(value); err == nil {
+		if value, err := validator.Validate(value); err == nil {
```
smyrman: The initial problem here is that we where overwriting the `value` functional parameter, so that the next iteration of the for-loop would use the value returned by the previous (failing) validator, which in most cases would be nil due to the returned error.

With this fix, we generate a _new_ (shadowing) value variable for each iteration. This variable is not passed on to the next iteration.

smyrman:
I will rebase one more time to put the bug-fixes in front of the feature, leaving all three commits with passing tests.

smyrman:
@rs, ready for final review. The two commits before the feature are (unrelated) bugfixes that where needed for the (updated) tests to pass, hench I recommend a **rebase and merge** so that they end up as separate commits on master :-)

smyrman:
Note tha the commit ordering on GitHub is a bit weird... Ir's the one with the green tick that's last. The git log on this is correct on the branch.

resource/index.go
```diff
@@ -83,21 +89,42 @@ func (r *index) GetResource(path string, parent *Resource) (*Resource, bool) {
...
+				return nil, err
 			}
-			return fmt.Errorf("%s%s%s", r.name, sep, err)
+		} else {
+			id = value
 		}
-	}
-	return nil
+
+		_, err = rsc.Get(context.TODO(), id)
```
rs: Why `TODO` over `Background`?

schema/anyof.go
```diff
@@ -5,23 +5,27 @@ import "errors"
...
-	return
+	return nil
 }
 
 // Validate ensures that at least one sub-validator validates.
 func (v AnyOf) Validate(value interface{}) (interface{}, error) {
 	for _, validator := range v {
-		var err error
-		if value, err = validator.Validate(value); err == nil {
+		if value, err := validator.Validate(value); err == nil {
```
rs: Good catch.

schema/field.go
```diff
@@ -83,18 +93,22 @@ type FieldSerializer interface {
...
 		}
 	} else if f.Validator != nil {
-		// Compile validator if it implements Compiler interface.
-		if c, ok := f.Validator.(Compiler); ok {
+		// Compile validator if it implements the ReferenceCompiler or Compiler interface.
+		if c, ok := f.Validator.(ReferenceCompiler); ok {
+			if err := c.Compile(rc); err != nil {
+				return fmt.Errorf(": %v", err)
+			}
+		} else if c, ok := f.Validator.(Compiler); ok {
```
rs: Wouldn't it be simpler to only have the `ReferenceCompiler` interface and remove the `Compiler` interface altogether? It is up to the type to use the `rc` or not.

Having both interfaces available is not clear on what happen if a given type implement both interfaces.

smyrman: It's not possible for one type to implement both interfaces since method loading is not legal in Go, and the methods have the same name

rs: Yes you are right. But having a single interface would be simpler and make sure we don’t forget to always check for both interfaces everywhere it’s needed.

smyrman: The `Compiler` interface is also used within the resource package, e.g. for the index.

I can check if it's feasible to move the `Compiler` interface to the `resource` package, and only use the new `ReferenceCompiler` interface within the `schema` package. It's an even bigger change, especially when it commes to updating tests, but should be mostly search/replace.

resource/index.go
smyrman: From the docs on `context.TODO()`:

> Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

From the docs on `context.Background()`:

> It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests.

FroI feel this case is closer to the description offered for `context.TODO()` than what's offered by `context.Background()`. I think if we where to use the _right_ context here, we should get it passed in from somewhere so that when using the `rest` package, we use the context created for a new request, and not a new context created later.

smyrman:
@rs, I have pushed *rename schema.ReferenceCompiler -> Compiler* as a new commit for easier review only.

When you are happy, I would like to do a manual rebase to fold this, as well as any other requested changes into the previous commit :-)

smyrman:
rebased

rs:
\o/

PR #101, opened 2017-03-27
Support date/time fields in gt, lt, gte & lte filter expressions

spiegela:
Should be fairly self explanatory.

I've looked through the code of the Mem, Elasticsearch and MongoDB adapters, and this appears that it should work fine. I tested all the way through with the Elasticsearch adapter, and it works great.

Let me know if there are any questions!

### DIFF
--- a/schema/query.go
+++ b/schema/query.go
@@ -64,25 +64,25 @@ type NotExist struct {
 // GreaterThan matches values that are greater than a specified value.
 type GreaterThan struct {
 	Field string
-	Value float64
+	Value interface{}
 }
 
 // GreaterOrEqual matches values that are greater than or equal to a specified value.
 type GreaterOrEqual struct {
 	Field string
-	Value float64
+	Value interface{}
 }
 
 // LowerThan matches values that are less than a specified value.
 type LowerThan struct {
 	Field string
-	Value float64
+	Value interface{}
 }
 
 // LowerOrEqual matches values that are less than or equal to a specified value.
 type LowerOrEqual struct {
 	Field string
-	Value float64
+	Value interface{}
 }
 
 // Regex matches values that match to a specified regular expression.
@@ -156,9 +156,13 @@ func validateQuery(q map[string]interface{}, validator Validator, parentKey stri
 			if parentKey == "" {
 				return nil, fmt.Errorf("%s can't be at first level", op)
 			}
-			n, ok := isNumber(exp)
-			if !ok {
-				return nil, fmt.Errorf("%s: value for %s must be a number", parentKey, op)
+			var v interface{}
+			if n, ok := isNumber(exp); ok {
+				v = n
+			} else if t, ok := isTime(exp); ok {
+				v = t
+			} else {
+				return nil, fmt.Errorf("%s: value for %s must be a number or time", parentKey, op)
 			}
 			if field := validator.GetField(parentKey); field != nil {
 				if field.Validator != nil {
@@ -174,13 +178,13 @@ func validateQuery(q map[string]interface{}, validator Validator, parentKey stri
 			}
 			switch op {
 			case "$gt":
-				queries = append(queries, GreaterThan{Field: parentKey, Value: n})
+				queries = append(queries, GreaterThan{Field: parentKey, Value: v})
 			case "$gte":
-				queries = append(queries, GreaterOrEqual{Field: parentKey, Value: n})
+				queries = append(queries, GreaterOrEqual{Field: parentKey, Value: v})
 			case "$lt":
-				queries = append(queries, LowerThan{Field: parentKey, Value: n})
+				queries = append(queries, LowerThan{Field: parentKey, Value: v})
 			case "$lte":
-				queries = append(queries, LowerOrEqual{Field: parentKey, Value: n})
+				queries = append(queries, LowerOrEqual{Field: parentKey, Value: v})
 			}
 		case "$in", "$nin":
 			op := key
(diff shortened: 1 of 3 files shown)


schema/query.go
```diff
@@ -354,25 +358,33 @@ func (e NotExist) Match(payload map[string]interface{}) bool {
 // Match implements Expression interface
 func (e GreaterThan) Match(payload map[string]interface{}) bool {
 	n, ok := isNumber(getField(payload, e.Field))
-	return ok && (n > e.Value)
+	if ok {
+		return ok && (n > e.Value.(float64))
+	}
+	t, ok := isTime(getField(payload, e.Field))
+	if ok {
```
spiegela: Thanks for the tips. Still getting used to the idioms in Go. In this block there are actually 2 type conversion checks (for both the field & filter values), so I'd like to get your feedback on the options to check each.  This stays more concise and inline with the original code:

```go
if t, ok := isTIme(getField(payload, e.Field)); ok {
    t2, ok2 := isTime(e.Value)
    return ok2 && t.After(t2)
}
```

This other seems more clear:

```go
if t, ok := isTIme(getField(payload, e.Field)); ok {
    if t2, ok2 := isTime(e.Value); ok2 {
        return t.After(t2)
    }
}
```

I'm leaning towards the 2nd option. Either is good by me.

schema/query.go
```diff
@@ -354,25 +358,33 @@ func (e NotExist) Match(payload map[string]interface{}) bool {
...
+		t2, ok := isTime(e.Value)
+		return ok && (t.After(t2))
+	}
+	return false
 }
 
 // Match implements Expression interface
 func (e GreaterOrEqual) Match(payload map[string]interface{}) bool {
```
spiegela: Whoops.  Should have pushed those as well.  I'll correct this once I finalize the other fixes.

schema/utils.go
```diff
@@ -35,6 +38,23 @@ func isNumber(n interface{}) (float64, bool) {
 	}
 }
 
+// isTime takes an interface as input, and returns a time.Time if the type is
+// a string and is formatted with RFC3339
+func isTime(t interface{}) (time.Time, bool) {
+	var t2 time.Time
+	var isTime bool
+	if tStr, ok := t.(string); ok {
+		t2, err := time.Parse(time.RFC3339, tStr)
```
rs: You should support all the formats defined in the [`time` validator](https://github.com/rs/rest-layer/blob/master/schema/time.go#L38).

The problem here, is that we don't have access to the schema, so we don't know if the field is supposed to be a time, and if user defines custom formats on the time field, these won't be supported here.

schema/query.go
```diff
@@ -64,25 +64,25 @@ type NotExist struct {
 // GreaterThan matches values that are greater than a specified value.
 type GreaterThan struct {
 	Field string
-	Value float64
+	Value interface{}
```
rs: Should be type `Value` instance of `interface{}`.

schema/query.go
rs: The `e.Value` should be casted to `time.Time` by `validateQuery` and `isTime` does not seem to check if the passed value is already a `time.Time`.

Also, as the data in the payload should be stored as casted by the validator (at least with mem storage which is the only storer to use those methods), shouldn't it be stored as a `time.Time` too?

Looks like the only case were those condition would work is when we compare a non time field with a time-like value.

In my understanding the condition should be written like this:

```go
func (e GreaterThan) Match(payload map[string]interface{}) bool {
	val := getField(payload, e.Field)
	if val, ok := isNumber(val); ok {
		if cond, ok := isNumber(e.Value) {
			return val > cond
		}
	}
	if val, ok := val.(time.Time); ok {
		if cond, ok := e.Value.(time.Time); ok {
			return val.After(cond)
		}
	}
	return false
}
```

schema/query_test.go
```diff
@@ -102,6 +111,10 @@ func TestParseQuery(t *testing.T) {
 	q, err = ParseQuery("{\"foo\": {\"$nin\": [\"bar\", \"baz\"]}}", s)
 	assert.NoError(t, err)
 	assert.Equal(t, Query{NotIn{Field: "foo", Values: []Value{"bar", "baz"}}}, q)
+	q, err = ParseQuery("{\"foo\": {\"$gte\": \"2009-05-20T10:23:34+00:00\"}}", s)
```
rs: We need a test that test against a field with a `Time` validator. Here your test is against a non-typed field which should not allow queries against time.

schema/query_test.go
```diff
@@ -158,7 +171,7 @@ func TestQueryInvalidType(t *testing.T) {
 	_, err = ParseQuery("{\"foo\": {\"$gt\": 1}}", s)
 	assert.EqualError(t, err, "foo: cannot apply $gt operation on a non numerical field")
 	_, err = ParseQuery("{\"bar\": {\"$gt\": \"1\"}}", s)
-	assert.EqualError(t, err, "bar: value for $gt must be a number")
+	assert.EqualError(t, err, "bar: value for $gt must be a number or time")
```
rs: It must be of the same type as the field. A query comparing time against number should not be possible.

schema/query.go
```diff
@@ -156,9 +156,13 @@ func validateQuery(q map[string]interface{}, validator Validator, parentKey stri
 			if parentKey == "" {
 				return nil, fmt.Errorf("%s can't be at first level", op)
 			}
-			n, ok := isNumber(exp)
-			if !ok {
-				return nil, fmt.Errorf("%s: value for %s must be a number", parentKey, op)
+			var v interface{}
+			if n, ok := isNumber(exp); ok {
+				v = n
+			} else if t, ok := isTime(exp); ok {
```
rs: You should only accept time condition if the validator is of type `Time` and number if of type `Float` or `Integer` (see on line 166 below).

Here you blindly accept time or number, and only on number fields and you refuse both on time fields.

schema/utils.go
rs: Actually we do, as this method should only be called from `validateQuery`. You can thus pass it the formats defined on the `Time` validator.

Dragomir-Ivanov:
Hi there,
What is the status of this merge request? What is needed to be polished for this to be merged? I need Date/Time filtering as well.

Thanks and regards

smyrman:
@Dragomir-Ivanov, I think this PR is pretty out of date. There is now a separate `query` package that didn't exist when the PR was started. There are also a few things to clean up to make sure it works as intended. If you want to have a try at it, it might just as good to start from scratch.

You could still look at this PR for inspiration.

smyrman:
Closing this one as it's out of date. Maybe a good solution could be implemented as part of #170.

PR #104, opened 2017-03-29
Use Replace instead of update for replacing item

muyadan:


rs:
Discussion about this issue can be found here: https://github.com/rs/rest-layer-mem/pull/4

PR #107, opened 2017-07-01
Improve routing test/bench and simplify index.GetResource

rs:


resource/index.go
```diff
@@ -55,33 +55,31 @@ func (r *index) Compile() error {
 //
 // If a parent is given and the path starts with a dot, the lookup is started at the
 // parent's location instead of root's.
-func (r *index) GetResource(path string, parent *Resource) (*Resource, bool) {
+func (r *index) GetResource(path string, parent *Resource) (sr *Resource, found bool) {
```
smyrman: I actually find the explicit double return more explicit and readable.

Although Go allows named returns, it may be error prone some times, especially with multiple return values. As it's less explicit, it's easier to (by mistake) return a pair of values that was not meant to be returned, and it's harder to read the code to find out which pair of variables are returned in different conditions.

 I tend to only use named returned parameters when it's necessary for technical reasons, such as setting an error in a deferred recover. The name can also serve as documentation when it's in the header, but even there it's only really needed if the semantic meaning of the return values can not be easily deduced from the method name, return types and go conventions.

Sorry for shooting this down, I just have done to a few to many mistakes myself with named return values when I started using Go (2010/2011), and they are still fresh in memory:-)

resource/index_bench_test.go
```diff
@@ -9,12 +9,48 @@ import (
...
+	tests := map[string]string{
+		"1st":      "route0001",
+		"100th":    "route0100",
+		"1000th":   "route1000",
+		"NotFound": "notfound",
 	}
+	for name, rsrc := range tests {
+		b.Run(name, func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				index.GetResource(rsrc, nil)
```
smyrman: No changes needed, this is just a FYI about a sub-test gatcha that applies to parallel/concurrent execution, in case you were not aware.

Using `rsrc` variable here _only_ works as long as the benchmarks here are not run in parallel.

If you where to add `b.Parallel()` above this line, the `rscrc` variables would be _shared_ among all the goroutines, i.e. it's reused for the next iteration of the for loop, leading to a data race condition. `name` is also reused in the next iteration, but as it's used only as an argument to Run, it happens before any new goroutine is started, so it's not a problem. To solve this in the case of parallelism we would need to capture the variable, e.g. like this:

```go
for name := range tests {
        rsrc := tests[name]  // the variable is declared within the iteration, and thus not reused on the next iteration.
```

Extract from https://blog.golang.org/subtests:
```go
func TestGroupedParallel(t *testing.T) {
    for _, tc := range testCases {
        tc := tc // capture range variable
        t.Run(tc.Name, func(t *testing.T) {
```
Again, there are no changes needed for this case, just something to be aware of.

resource/index_bench_test.go
```diff
@@ -9,12 +9,48 @@ import (
...
+	b.Run("TwoLevels", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			_, _ = r.GetResource("foo.bar", nil)
+		}
+	})
+	b.Run("ThreeLevels", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			_, _ = r.GetResource("foo.bar.baz", nil)
+		}
+	})
```
smyrman: I guess you could use a `map [string]string` and for loop here as well, making the code more DRY.

resource/index.go
```diff
@@ -55,33 +55,31 @@ func (r *index) Compile() error {
...
 	resources := r.resources
 	if len(path) > 0 && path[0] == '.' {
 		if parent == nil {
 			// If field starts with a dot and no parent is given, fail the lookup
-			return nil, false
+			return
```
smyrman: Apparently, this is called a "naked return".

Quoting https://tour.golang.org/basics/7:

> Naked return statements should be used only in short functions, as with the example shown here. They can harm readability in longer functions.

PR #108, opened 2017-07-03
New query package with custom parser

rs:
This PR introduces a new query parser that is fully compatible with current query syntax, but also introduce alternative MongoDB like syntax with quote omitted for keys (i.e.: `{foo: {$in: ["a", "b"]}}`). This parser is also more efficient because it does everything in a single pass instead of having to first parse JSON then transform the JSON result into a `Query`.

This PR does also introduce a breaking change: all the query code is moved into its dedicated `query` package. I did that because I think having `And`, `In` etc. mixed with other `schema` types is confusing. I will also push changes to `Storer` to reflect this change and send PR to the public third party `Storer` implementations.


smyrman:
> ll the query code is moved into its dedicated query package. I did that because I think having And, In etc. mixed with other schema types is confusing.ll the query code is moved into its dedicated query package. I did that because I think having And, In etc. mixed with other schema types is confusing.

Sweet :-)

schema/query/query.go
```diff
@@ -0,0 +1,383 @@
...
+	cmdOr            = "$or"
+	cmdExists        = "$exists"
+	cmdIn            = "$in"
+	cmdNotIn         = "$nin"
+	cmdNotEqual      = "$ne"
+	cmdLowerThan     = "$lt"
+	cmdLowerOrEqual  = "$lte"
+	cmdGeaterThan    = "$gt"
+	cmdGeaterOrEqual = "$gte"
+	cmdRegex         = "$regex"
```
smyrman: Since this names are internal, it doesn't matter to much, but it's probably better to prefix them 'op' for operator rather than cmd for command;

https://docs.mongodb.com/manual/reference/operator/query/

schema/query/parse_test.go
```diff
@@ -0,0 +1,348 @@
...
+			Query{},
+			errors.New("char 1: $in: invalid placement"),
+		},
+		{
+			`{"$regex": "someregexpression"}`,
+			Query{},
+			errors.New("char 1: $regex: invalid placement"),
+		},
+	}
+	for _, tt := range tests {
```
smyrman: Should this use sub-tests for better listing of the results, and maybe do them in parallel?


```go
 +	for i := range tests {
 +		tt := tests[i]   // capture sub-test.
 +		t.Run(strings.Replace(tt.query, " ", "", -1), func(t *testing.T) {
 +			t.Parallel()
 +			got, err := Parse(tt.query)
 +			if !reflect.DeepEqual(err, tt.err) {
...
 + 		})
 +	}
```

schema/query/query_test.go
```diff
@@ -0,0 +1,151 @@
...
+		},
+		{
+			`{"$and": [{"foo": "bar"}, {"foo": "baz"}]}`, []test{
+				{map[string]interface{}{"foo": "bar"}, false},
+				{map[string]interface{}{"foo": "baz"}, false},
+				{map[string]interface{}{"bar": float64(1)}, false},
+			},
+		},
+	}
+	for _, tt := range tests {
```
smyrman: Same here, rely on sub-tests for better result printing?

The sub-test name can be tt.query with spaces stripped out to avoid repeating yourself. The reason for stripping out spaces is just that it looks slightly better in the listing, as the test package would translate spaces to underscore if they are not removed, presumably to prevent tripping up potential test result parsers.

schema/query/parse_test.go
rs: Good point, I'll try that.

rs:
@smyrman what do you think about the backward incompatible package split? Do you also feel it's needed?

smyrman:
> @smyrman what do you think about the backward incompatible package split? Do you also feel it's needed?

Actually, one of the first thing that annoyed me a bit when I first read the rest-layer documentation, is how mixed the query types and schema types where in the package documentation -- especially as they are just ordered alphabetically. Therefore, I am all for this change.

If we _want_ to keep backwards-compatiblity for a while, Go 1.9 launches type aliases which could be cool to try out, but otherwise I feel the best way of handling backwards compatibility for now is to just tag some versions every now and then to make it easier for the users to maintain their code. This also includes tagging the _storage_ backends with a version, so that it's possible for the users to require a particular version of both without having to resolve to Git hashes.

README.md
```diff
@@ -845,31 +845,31 @@ U              ungreedy: swap meaning of x* and x*?, x+ and x+?, etc (default fa
 For example the following regular expression would match any document with a field `type` and its value `rest-layer`.
 
 ```json
```
smyrman: Try js or JavaScript in stead of json to get rid of highlight errors when the key is not quoted. This is basically JS styled syntax.

smyrman:
Since we where talking about tagging, it could maybe be good to tag all the repos at their current state with a v0.x.y release before merging this. I guess it's not too many known _bugs_ right now?

Then we can wait with tagging for a while until we do a few more of those breaking changes?

rs:
I though about Go 1.9 type alias, but as it is a not yet released version of Go, it is hard to imagine that it will be useful for anybody until hold this PR for a few months. Let's break it fast and clean.

PR #115, opened 2017-07-14
Migrate resource.Lookup into query.Query (breaking change)

rs:
Having pagination and projection in the `resource` package (`Lookup` type) and the query in the `schema/query` package feels odd. If we want the `schema` and the `query` package to be really idependant, all the tools to perform a query should be in there.

This change moves all the `resource.Lookup` type into the `query.Query`. The previous `query.Query` type become `query.Predicate` and is now a field of the `query.Query` type. The former `resource.Lookup` selector field becomes the `Projection` field of the `query.Query` type and the pagination (`offset` and `limit`) are now embedded into the query instead of being passed as separate parameter to the `Storer`.

This change breaks compatibility with `Storers`. They all now have to replace `resource.Lookup`, `offset` and `limit` parameter by this new `query.Query` parameter.

One cool thing about this, is that the previously private `connection` validator to handle connection in the `resource` package is now public and is able to handle connections and references without depending on the `resource` package. This will helps people wanting to use `schema` and `schema/query` independantly from the rest of the project. This does also make the code cleaner and easier to understand.

I am still not happy with the way we handle sub-request for connection and reference resolution. The coalescing code is quite messy, and there is no good test coverage. The test coverage for the projection is not great easier. Next commit will refactor the sub-request code and improve tests.

Ultimately, tell me what you think @smyrman, I'd like to get rid of testify for test, and make pure Go tests with mainly table tests.


schema/query/query.go
```diff
@@ -0,0 +1,72 @@
...
+//
+// Example:
+//
+//   New("foo{bar},baz:b", `{baz: "bar"}`, "foo.bar,-baz", Page(1, 10, 0))
+//
+// Select items with the foo field equal to bar, including only the foo.bar and
+// baz fields in the result set, with the baz field aliased to b. The result is
+// then sorted by the foo.bar field ascending and baz descending. The result is
+// windowed on page 1 with 10 items per page, skiping no result.
+func New(projection, predicate, sort string, window *Window) (*Query, error) {
```
smyrman: I don't see this function being called anywhere. Is it meant to be called within e.g. the rest package?

rs: You're right, in all cases I need to incrementally build the query package. It might be better to have `New` creating an empty query and let the caller set the public fields.

smyrman: If it is not going to do anything other than `&query.Query{}`, then it might be best to drop it from the public API for now, and add it in later. The naming of this function is perfect idiomatic, but it's also idiomatic to not add a creator function if it doesn't do anything.

If we want to add a `New` later, maybe we see that it's convenient to pass it some positional parameter(s) after all, and we could add it without breaking changes.

rs:
@smyrman I refactored the reference resolution code. Tell me what you think. I will work on the tests next.

schema/query/projection_evaluator_test.go
```diff
@@ -0,0 +1,110 @@
...
+	assert.EqualError(t, err, "simple: field as no children")
+	assert.Nil(t, p)
+	// Deep field lookup on a field with invalid payload (no dict)
+	pr = Projection{{Name: "parent", Children: Projection{{Name: "child"}}}}
+	p, err = pr.Eval(ctx, map[string]interface{}{"parent": "value"}, s, nil)
+	assert.EqualError(t, err, "parent: invalid value: not a dict")
+	assert.Nil(t, p)
+}
+
+// func TestLookupApplySelector(t *testing.T) {
```
smyrman: Is this test meant to be rewritten or removed?

graphql/query.go
```diff
@@ -78,15 +79,11 @@ var listArgs = graphql.FieldConfigArgument{
 	},
 }
 
-func listParamResolver(r *resource.Resource, p graphql.ResolveParams, params url.Values) (lookup *resource.Lookup, offset int, limit int, err error) {
+func listParamResolver(r *resource.Resource, p graphql.ResolveParams, params url.Values) (q *query.Query, err error) {
```
smyrman: FYI: For _later_ improvement only.

As per previous discussion on `New()`, I think the goal of a New function should be to reduce the code, error checking and validation necessary in this an similar functions. I don't know if it would help, but here is an example from the top of my head:

`query` package:
```go
// Parser should extract the raw string values and
// return them for further validation by e.g. New.
type Parser interface{
        Projection() string
        Filter() string
        Sort() string
        Window() *Window
}

// Parse returns a new Query based on the raw values extracted from p. 
Parse(p Parser) (*Query, error) {...}
```
In this package (and similar in the `rest` package):
````go
type QueryParser struct{}

NewQueryParser(p graphql.ResolveParams, params url.Values) (*QueryParser, error)
```

smyrman:
> @smyrman I refactored the reference resolution code. Tell me what you think. I will work on the tests next.

I might need some time to fully understand it first :-)

schema/query/projection_evaluator.go
```diff
@@ -50,23 +53,31 @@ func evalProjection(ctx context.Context, p Projection, payload map[string]interf
...
+					// Execute sub-request in batch
+					q := &Query{Predicate: Predicate{Equal{Field: "id", Value: val}}}
+					rbr.request(ref.Path, q, func(payloads []map[string]interface{}, validator schema.Validator) error {
+						resMu.Lock()
+						defer resMu.Unlock()
+						if len(payloads) != 1 {
+							res[name] = nil
+							return nil
+						}
+						payload, err := evalProjection(ctx, pf.Children, payloads[0], validator, rbr)
```
smyrman: Is resMu meant to be locked _while_ this recursive call is ongoing?

schema/query/projection_evaluator.go
```diff
@@ -50,23 +53,31 @@ func evalProjection(ctx context.Context, p Projection, payload map[string]interf
...
-						field:   pf,
-						path:    ref.Path,
-						id:      val,
-						def:     def,
-					}
+					// Execute sub-request in batch
+					q := &Query{Predicate: Predicate{Equal{Field: "id", Value: val}}}
+					rbr.request(ref.Path, q, func(payloads []map[string]interface{}, validator schema.Validator) error {
+						resMu.Lock()
+						defer resMu.Unlock()
```
smyrman: should resMu really just be locked while you read out res[name]? Does it make sense to make it a RWMutex?

schema/query/projection_evaluator.go
```diff
@@ -79,33 +90,31 @@ func evalProjection(ctx context.Context, p Projection, payload map[string]interf
...
 					return nil, err
 				}
-				res[name] = handler
+				rbr.request(ref.Path, q, func(payloads []map[string]interface{}, validator schema.Validator) (err error) {
+					for i := range payloads {
+						if payloads[i], err = evalProjection(ctx, pf.Children, payloads[i], validator, rbr); err != nil {
+							return fmt.Errorf("%s: error applying projection on sub-resource item #%d: %v", pf.Name, i, err)
+						}
+					}
+					resMu.Lock()
```
smyrman: should this be locked _while_ resolveFieldHandler is running, or only when res[name] is set?

schema/query/resolver.go
```diff
@@ -0,0 +1,143 @@
...
+
+type referenceResponseHandler func(payloads []map[string]interface{}, validator schema.Validator) error
+
+type referenceBatchResolver struct {
+	mu       sync.Mutex
+	requests []referenceRequest
+}
+
+func (rbr *referenceBatchResolver) request(resourcePath string, q *Query, handler referenceResponseHandler) {
+	rbr.mu.Lock()
```
smyrman: Only one `request` might run at the time per rbr instance - is that intentional?

schema/query/projection_evaluator.go
smyrman: Like this?
```go
if v, err := resolveFieldHandler(ctx, pf, def, payloads); err == nil {
    resMu.Lock()
    res[name] = v
    resMu.Unlock()
} else {
    return fmt.Errorf("%s: error resolving field handler on sub-resource: %v", name, err)
}
return nil
```

schema/query/projection_evaluator.go
```diff
@@ -4,24 +4,27 @@ import (
...
+		// Execute the batched reference resolutions.
+		err = rbr.execute(ctx, resolver)
 	}
 	return payload, err
 }
 
-func evalProjection(ctx context.Context, p Projection, payload map[string]interface{}, validator schema.Validator, resolver ReferenceResolver) (map[string]interface{}, error) {
+func evalProjection(ctx context.Context, p Projection, payload map[string]interface{}, validator schema.Validator, rbr *referenceBatchResolver) (map[string]interface{}, error) {
 	res := map[string]interface{}{}
+	resMu := sync.Mutex{} // XXX use sync.Map once Go 1.9 is out
```
smyrman: Maybe.  Although `sync.Map` is very cool, it's a very complex implementation (from what little I have seen so far`), that has some extra allocates, e.g. to achieve lock-free reads.

Then functions under evalProjection seams to be mostly _writing_ to the map. If the mutex can be used right, it's not _necessarily_ true that a `sync.Map` will be faster, although it's somewhat easier to not screw it up.  It might be good with a benchmark to help make such a decision.

schema/query/projection_evaluator_test.go
rs: Yes, I will rewrites tests.

schema/query/projection_evaluator.go
rs: No, you are right. It won’t deadlock though.

schema/query/resolver.go
rs: No, one request can be "scheduled" at a time. Then the execution is parallel.

schema/query/projection_evaluator.go
rs: Oh I though I removed this comment. I looked at the type then went to the same conclusion.

schema/query/projection_evaluator.go
```diff
@@ -0,0 +1,170 @@
...
+
+	"sync"
+
+	"github.com/rs/rest-layer/schema"
+)
+
+// Eval evaluate the projection on the given payload with the help of the
+// validator. The resolver is used to fetch payload of references outside of the
+// provided payload.
+func (p Projection) Eval(ctx context.Context, payload map[string]interface{}, validator schema.Validator, resolver ReferenceResolver) (map[string]interface{}, error) {
```
smyrman: > Looks like the resolver parameter is not used? Should it be passed to rbr?

 IT is, on line 20 🤦‍♂️

rest/util.go
```diff
@@ -175,17 +177,25 @@ func checkIntegrityRequest(r *http.Request, original *resource.Item) *Error {
 	return nil
 }
 
-func getReferenceResolver(ctx context.Context, r *resource.Resource) resource.ReferenceResolver {
-	return func(path string) (*resource.Resource, error) {
+func getReferenceResolver(r *resource.Resource) query.ReferenceResolver {
+	return func(ctx context.Context, path string, q *query.Query) ([]map[string]interface{}, schema.Validator, error) {
 		router, ok := IndexFromContext(ctx)
```
rs: Agreed. We might think about something in a separate PR.

schema/query/projection_evaluator.go
rs: I'm not really satisfied by this API either. It is quite confusing that you have to pass a validator and a function that returns a payload AND a validator. I would rather request an interface as you described. Isn't your first example's `ReferenceResolver()` method supposed to return a `Resource` interface and not a `ReferenceResolver`?

smyrman: > Isn't your first example's ReferenceResolver() method supposed to return a Resource interface and not a ReferenceResolver?

Not according to my original thinking on the spot, but that might be a better solution? I suppose then the interface also needs a lookup method for the Payload.

rs: Another idea:

```go
type Items interface {
    Payloads() []map[string]interface{}
}

type Resource interface {
    Find(ctx context.Context, query query.Query) (Items, error)
    MultiGet(ctx context.Context, ids []interface{}) (Items, error)
    GetSubResource(path string) (Resource, error)
    Validator() schema.Validator
}

func (p Projection) Eval(ctx context.Context, payload map[string]interface{}, rsc Resource) (map[string]interface{}, error) {}
```

This solves another issue `MultiGet`. I currently "hack" the find method to simulate a multi-get, but as commented in the code, this is dangerous because it assumes the `id` in the payloads can be compared using equality with the queried ids, which might not always be the case.

With those interface, it should be easy to make the current `resource.Resource` type compatible without intermediary type, which is better for performance.

smyrman: If the interface is to return payloads, the interface name is probably better called "Items", or to use better query terminology a "Selection":

```go
type Selection interface{
        Items() []map[string]interface{}
        ResolveReference(ctx context.Context, path string, *Query) Selection
        Validator() shema.Validator
}
```

However it's designed, it would be pretty cool if the API allows a future possibility to provide custom RefrenceResolvers, e.g. to compose connections to _other_ REST APIs, which might come in handy for some micro-services.

smyrman: And then for projection I suppose:
```go
+func (p Projection) Eval(ctx context.Context, Selection) (map[string]interface{}, error)
```

rs: Please check the last commit.

smyrman: > Another idea:
I didn't see your comment when I posted my last-two comments btw. I will have a look at the last commit.

rest/resource.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/rs/rest-layer/resource"
+	"github.com/rs/rest-layer/schema/query"
+)
+
+type restResource struct {
+	*resource.Resource
+}
+
+// GetSubResource implements query.Resource interface.
```
smyrman: // Find implements ...

rest/resource.go
```diff
@@ -0,0 +1,53 @@
...
+		return nil, err
+	}
+	payloads := make([]map[string]interface{}, 0, len(itemList.Items))
+	for _, i := range itemList.Items {
+		payloads = append(payloads, i.Payload)
+	}
+	return payloads, nil
+}
+
+// GetSubResource implements query.Resource interface.
```
smyrman: // MultiGet ...

rest/method_get.go
```diff
@@ -43,7 +43,7 @@ func listGet(ctx context.Context, r *http.Request, route *RouteMatch) (status in
 		list.Offset = win.Offset
 	}
 	for _, item := range list.Items {
-		item.Payload, err = q.Projection.Eval(ctx, item.Payload, rsc.Validator(), getReferenceResolver(rsc))
+		item.Payload, err = q.Projection.Eval(ctx, item.Payload, restResource{rsc})
```
smyrman: Just trying to compare this against the `query.Selection` idea, not trying to state which one is better:

With the `Selection` idea, this would be (without the for-loop on line 45):

```go
payloads, err = q.Projection.Eval(ctx, restSelection{resource: rsc, items: list.Items})
if err != nil {...}
for i := payloads {
    list.Items[i] = payloads[i]
}
```

rest/resource.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	payloads := make([]map[string]interface{}, 0, len(items))
+	for _, i := range items {
+		payloads = append(payloads, i.Payload)
+	}
+	return payloads, nil
+}
+
+// GetSubResource implements query.Resource interface.
+func (r restResource) GetSubResource(ctx context.Context, path string) (query.Resource, error) {
```
smyrman: Rename to SubResource.

It's more idiomatic in Go to drop "Get" from getter methods. Or to be more accurate, it's more idiomatic to let names be as short as they can be while still making perfect sense. Just "SubResource" is shorter, but still reads equally well, so therefore it's preferred.

- https://golang.org/doc/effective_go.html#Getters

This does not mean that the _word_ Get is banned in Go (It's perfectly well used when referring to e.g. the HTTP GET Method), just that it serves no purpose in most getter methods.

rest/method_get.go
smyrman: After some reflectionI think `query.Resource` is an overall better idea, and easier to understand.

It _might_ also be that with some further refactoring of resource.Resource, it _could_ be made to implement `query.Resource`, which would simply usage. The `resource.Item` type would plausibly have to be dropped (or moved) to achieve this.

smyrman: For now I think `query.Resource` is an overall better idea btw.

rs:
@smyrman: ready to merge, please have a final review.

resource/storage.go
```diff
@@ -8,23 +8,27 @@ import (
...
 	// the user explicitly request the total.
 	//
-	// The whole lookup query must be treated. If a query operation is not
-	// implemented by the storage handler, a resource.ErrNotImplemented must be
-	// returned.
+	// The whole query must be treated. If a query predicate operation or sort
+	// is not implemented by the storage handler, a resource.ErrNotImplemented
+	// must be returned.
+	//
+	// A storer must ignore the Projection part of the query and always return
```
smyrman: Alright, I guess this makes sense initially.

resource/storage.go
```diff
@@ -246,25 +250,25 @@ func (s storageWrapper) Delete(ctx context.Context, item *Item) (err error) {
...
 	}
 	if ctx.Err() != nil {
 		return -1, ctx.Err()
 	}
 	if c, ok := s.Storer.(Counter); ok {
-		return c.Count(ctx, lookup)
+		return c.Count(ctx, q)
 	}
 	return -1, ErrNotImplemented
 }
```
smyrman: ACK

rest/doc.go
```diff
@@ -2,6 +2,7 @@
 Package rest is a `net/http` handler responsible for HTTP RESTful implementation
 for the REST Layer framework.
 
-See http://github.com/rs/rest-layer for full REST Layer documentation.
+This package is part of the rest-layer project. See http://rest-layer.io for
+full REST Layer documentation.
 */
 package rest
```
smyrman: ACK

rest/method_delete.go
```diff
@@ -8,11 +8,11 @@ import (
...
-	lookup, e := route.Lookup()
+	q, e := route.Query()
 	if e != nil {
 		return e.Code, nil, e
 	}
-	total, err := route.Resource().Clear(ctx, lookup)
+	total, err := route.Resource().Clear(ctx, q)
 	if err != nil {
 		e = NewError(err)
 		return e.Code, nil, e
```
smyrman: ACK

rest/method_delete_test.go
```diff
@@ -73,7 +74,7 @@ func TestHandlerDeleteListFilter(t *testing.T) {
 	assert.Equal(t, http.Header{"X-Total": []string{"2"}}, headers)
 	assert.Nil(t, body)
 
-	l, err := s.Find(context.TODO(), resource.NewLookup(), 0, -1)
+	l, err := s.Find(context.TODO(), &query.Query{})
 	assert.NoError(t, err)
 	if assert.Len(t, l.Items, 3) {
 		assert.Equal(t, "3", l.Items[0].ID)
```
smyrman: ACK

rest/method_get.go
```diff
@@ -58,24 +24,26 @@ func listGet(ctx context.Context, r *http.Request, route *RouteMatch) (status in
...
-	list.Offset = offset
+	if win := q.Window; win != nil && win.Offset > 0 {
+		list.Offset = win.Offset
+	}
 	for _, item := range list.Items {
-		item.Payload, err = lookup.ApplySelector(ctx, rsrc.Validator(), item.Payload, getReferenceResolver(ctx, rsrc))
+		item.Payload, err = q.Projection.Eval(ctx, item.Payload, restResource{rsc})
 		if err != nil {
 			e = NewError(err)
 			return e.Code, nil, e
```
smyrman: ACK

rest/method_get_test.go
```diff
@@ -271,6 +275,7 @@ func TestHandlerGetListFieldHandlerError(t *testing.T) {
 	}, s, resource.DefaultConf)
 	r, _ := http.NewRequest("GET", "/test", nil)
 	rm := &RouteMatch{
+		Method: "GET",
 		ResourcePath: []*ResourcePathComponent{
 			&ResourcePathComponent{
 				Name:     "test",
```
smyrman: ACK

rest/method_item_delete.go
```diff
@@ -3,15 +3,18 @@ package rest
...
+	q, e := route.Query()
 	if e != nil {
 		return e.Code, nil, e
 	}
-	l, err := route.Resource().Find(ctx, lookup, 0, 1)
+	q.Window = &query.Window{Limit: 1}
+	l, err := route.Resource().Find(ctx, q)
 	if err != nil {
 		e = NewError(err)
 		return e.Code, nil, e
```
smyrman: ACK

rest/method_item_delete_test.go
```diff
@@ -142,7 +143,7 @@ func TestHandlerDeleteItemWrongEtag(t *testing.T) {
 		assert.Equal(t, "Precondition Failed", err.Message)
 	}
 
-	l, err := s.Find(context.TODO(), resource.NewLookup(), 0, -1)
+	l, err := s.Find(context.TODO(), &query.Query{})
 	assert.NoError(t, err)
 	assert.Len(t, l.Items, 1)
 }
```
smyrman: ACK

rest/method_item_get.go
```diff
@@ -35,7 +38,7 @@ func itemGet(ctx context.Context, r *http.Request, route *RouteMatch) (status in
 			return 304, nil, nil
 		}
 	}
-	item.Payload, err = lookup.ApplySelector(ctx, rsrc.Validator(), item.Payload, getReferenceResolver(ctx, rsrc))
+	item.Payload, err = q.Projection.Eval(ctx, item.Payload, restResource{rsrc})
 	if err != nil {
 		e = NewError(err)
 		return e.Code, nil, e
```
smyrman: ACK

rest/method_item_get_test.go
```diff
@@ -68,7 +68,7 @@ func TestHandlerGetItemInvalidLookupSort(t *testing.T) {
 	}
 }
 
-func TestHandlerGetItemInvalidLookupFilter(t *testing.T) {
+func TestHandlerGetItemInvalidQueryFilter(t *testing.T) {
 	index := resource.NewIndex()
 	test := index.Bind("test", schema.Schema{}, nil, resource.DefaultConf)
 	r, _ := http.NewRequest("GET", "/test", nil)
```
smyrman: ACK

rest/method_item_patch.go
```diff
@@ -62,8 +64,8 @@ func itemPatch(ctx context.Context, r *http.Request, route *RouteMatch) (status
 		e = NewError(err)
 		return e.Code, nil, e
 	}
-	// Apply selector so response gets the same format as read requests.
-	item.Payload, err = lookup.ApplySelector(ctx, rsrc.Validator(), item.Payload, getReferenceResolver(ctx, rsrc))
+	// Evaluate projection so response gets the same format as read requests.
+	item.Payload, err = q.Projection.Eval(ctx, item.Payload, restResource{rsrc})
 	if err != nil {
 		e = NewError(err)
 		return e.Code, nil, e
```
smyrman: ACK

rest/method_item_patch_test.go
```diff
@@ -51,7 +54,7 @@ func TestHandlerPatchItemBadPayload(t *testing.T) {
 	}
 }
 
-func TestHandlerPatchItemInvalidLookupFields(t *testing.T) {
+func TestHandlerPatchItemInvalidQueryFields(t *testing.T) {
 	index := resource.NewIndex()
 	test := index.Bind("test", schema.Schema{}, nil, resource.DefaultConf)
 	r, _ := http.NewRequest("PATCH", "/test/2", bytes.NewBufferString("{}"))
```
smyrman: ACK

rest/method_item_put.go
```diff
@@ -94,8 +96,8 @@ func itemPut(ctx context.Context, r *http.Request, route *RouteMatch) (status in
 			return e.Code, nil, e
 		}
 	}
-	// Apply selector so response gets the same format as read requests.
-	item.Payload, err = lookup.ApplySelector(ctx, rsrc.Validator(), item.Payload, getReferenceResolver(ctx, rsrc))
+	// Evaluate projection so response gets the same format as read requests.
+	item.Payload, err = q.Projection.Eval(ctx, item.Payload, restResource{rsrc})
 	if err != nil {
 		e = NewError(err)
 		return e.Code, nil, e
```
smyrman: ACK

rest/method_item_put_test.go
```diff
@@ -46,7 +49,7 @@ func TestHandlerPutItemBadPayload(t *testing.T) {
 	}
 }
 
-func TestHandlerPutItemInvalidLookupFields(t *testing.T) {
+func TestHandlerPutItemInvalidQueryFields(t *testing.T) {
 	index := resource.NewIndex()
 	test := index.Bind("test", schema.Schema{}, nil, resource.DefaultConf)
 	r, _ := http.NewRequest("PUT", "/test/2", bytes.NewBufferString("{}"))
```
smyrman: ACK

rest/method_post.go
```diff
@@ -39,8 +39,8 @@ func listPost(ctx context.Context, r *http.Request, route *RouteMatch) (status i
 		e = NewError(err)
 		return e.Code, nil, e
 	}
-	// Apply selector so response gets the same format as read requests.
-	item.Payload, err = lookup.ApplySelector(ctx, rsrc.Validator(), item.Payload, getReferenceResolver(ctx, rsrc))
+	// Evaluate projection so response gets the same format as read requests.
+	item.Payload, err = q.Projection.Eval(ctx, item.Payload, restResource{rsrc})
 	if err != nil {
 		e = NewError(err)
 		return e.Code, nil, e
```
smyrman: ACK

rest/method_post_test.go
```diff
@@ -33,12 +33,15 @@ func TestHandlerPostList(t *testing.T) {
...
+				}
 				s, ok := vars.Storers["foo"]
 				if !assert.True(t, ok) {
 					return
 				}
-				l, err := s.Find(context.TODO(), lkp, 0, 1)
+				l, err := s.Find(context.TODO(), q)
 				assert.NoError(t, err)
 				if assert.Len(t, l.Items, 1) {
 					assert.Equal(t, map[string]interface{}{"id": "1", "foo": "bar"}, l.Items[0].Payload)
```
smyrman: ACK

rest/method_post_test.go
```diff
@@ -60,7 +63,7 @@ func TestHandlerPostList(t *testing.T) {
 				"message": "Malformed body: invalid character 'i' looking for beginning of object key string"
 			}`,
 		},
-		"InvalIDLookupFields": {
+		"InvalIDQueryFields": {
 			Init: func() *requestTestVars {
 				index := resource.NewIndex()
 				index.Bind("test", schema.Schema{}, nil, resource.DefaultConf)
```
smyrman: ACK

rest/resource.go
```diff
@@ -0,0 +1,53 @@
...
+	router, ok := IndexFromContext(ctx)
+	if !ok {
+		return restResource{}, errors.New("router not available in context")
+	}
+	rsc, found := router.GetResource(path, r.Resource)
+	if !found {
+		return restResource{}, fmt.Errorf("invalid resource reference: %s", path)
+	}
+	return restResource{rsc}, nil
+}
```
smyrman: ACK

rest/resource_path.go
```diff
@@ -128,7 +128,7 @@ func (p ResourcePath) ParentsExist(ctx context.Context) error {
 			}
 		}(i)
 		// Push the resource field=value for the next hops.
-		q = append(q, query.Equal{Field: p[i].Field, Value: p[i].Value})
+		predicate = append(predicate, query.Equal{Field: p[i].Field, Value: p[i].Value})
 	}
 	// Fail on first error.
 	for i := 0; i < parents; i++ {
```
smyrman: ACK

rest/resource_test.go
```diff
@@ -0,0 +1,22 @@
...
+	index := resource.NewIndex()
+	index.Bind("foo", schema.Schema{}, nil, resource.DefaultConf)
+	ctx := context.Background()
+	rsc := restResource{}
+	_, err := rsc.SubResource(ctx, "bar")
+	assert.EqualError(t, err, "router not available in context")
+	ctx = contextWithIndex(ctx, index)
+	_, err = rsc.SubResource(ctx, "bar")
+	assert.EqualError(t, err, "invalid resource reference: bar")
+}
```
smyrman: ACK

rest/routing.go
```diff
@@ -180,36 +180,83 @@ func (r *RouteMatch) ResourceID() interface{} {
...
+		if err != nil {
 			return nil, &Error{422, fmt.Sprintf("Invalid `fields` parameter: %s", err), nil}
 		}
+		q.Projection = p
 	}
-	return l, nil
+	return q, nil
 }
 
 // Release releases the route so it can be reused.
```
smyrman: ACK

rest/routing_test.go
```diff
@@ -208,9 +208,9 @@ func TestRoutePathParentsExists(t *testing.T) {
...
 		assert.Len(t, h.queries, 2)
 		// query on /foo/1234
-		assert.Contains(t, h.queries, query.Query{query.Equal{Field: "id", Value: "1234"}})
+		assert.Contains(t, h.queries, query.Query{Predicate: query.Predicate{query.Equal{Field: "id", Value: "1234"}}})
 		// query on /bar/5678 with foo/1234 context
-		assert.Contains(t, h.queries, query.Query{query.Equal{Field: "f", Value: "1234"}, query.Equal{Field: "id", Value: "5678"}})
+		assert.Contains(t, h.queries, query.Query{Predicate: query.Predicate{query.Equal{Field: "f", Value: "1234"}, query.Equal{Field: "id", Value: "5678"}}})
 	}
 
 	route = newRoute("GET")
```
smyrman: ACK

rest/util.go
```diff
@@ -175,20 +174,6 @@ func checkIntegrityRequest(r *http.Request, original *resource.Item) *Error {
...
-		if !found {
-			return nil, fmt.Errorf("invalid resource reference: %s", path)
-		}
-		return rsrc, nil
-	}
-}
-
 func logErrorf(ctx context.Context, format string, a ...interface{}) {
 	if resource.Logger != nil {
 		resource.Logger(ctx, resource.LogLevelError, fmt.Sprintf(format, a...), nil)
```
smyrman: ACK

rest/util_test.go
```diff
@@ -240,18 +238,3 @@ func TestRequestCheckIntegrityModifiedDateMatch(t *testing.T) {
...
-	_, err := getReferenceResolver(ctx, nil)("bar")
-	assert.EqualError(t, err, "router not available in context")
-	ctx = contextWithIndex(ctx, index)
-	_, err = getReferenceResolver(ctx, nil)("bar")
-	assert.EqualError(t, err, "invalid resource reference: bar")
-	r, err := getReferenceResolver(ctx, nil)("foo")
-	if assert.NoError(t, err) {
-		assert.Equal(t, "foo", r.Name())
-	}
-}
```
smyrman: ACK

schema/connection.go
```diff
@@ -0,0 +1,18 @@
...
+
+// Connection is a dummy validator to define a weak connection to another
+// schema. The query.Projection will treat this validator as an external
+// resource, and generate a sub-request to fetch the sub-payload.
+type Connection struct {
+	Path string
+}
+
+func (v *Connection) Compile(rc ReferenceChecker) (err error) {
+	// Nothing to compile, implemented to force Connection on pointer.
```
smyrman: This is hacky, just move Validate on pointer as that will do the same, but compile time.

schema/connection.go
```diff
@@ -0,0 +1,18 @@
...
+	return nil
+}
+
+func (v Connection) Validate(value interface{}) (interface{}, error) {
```
smyrman: Missing doc-string, move to pointer and remove Compile

schema/dependency.go
```diff
@@ -4,15 +4,15 @@ import (
...
 
-// Q is deprecated, use query.MustParse instead.
-func Q() Query {
-	panic("schema.Q is deprecated, please use query.MustParse instead")
+// Q is deprecated, use query.MustParsePredicate instead.
+func Q() Predicate {
+	panic("schema.Q is deprecated, please use query.MustParsePredicate instead")
 }
 
 // compileDependencies recursively compiles all field.Dependency against the
```
smyrman: ACK

schema/doc.go
```diff
@@ -1,6 +1,7 @@
 /*
 Package schema provides a validation framework for the API resources.
 
-See http://github.com/rs/rest-layer for full REST Layer documentation.
+This package is part of the rest-layer project. See http://rest-layer.io for
+full REST Layer documentation.
 */
 package schema
```
smyrman: ACK

schema/field.go
```diff
@@ -48,9 +48,9 @@ type Field struct {
...
 	Validator FieldValidator
-	// Dependency rejects the field if the schema query doesn't match the document.
-	// Use query.MustParse(`{"field": "value"}`) to populate this field.
-	Dependency Query
+	// Dependency rejects the field if the schema predicate doesn't match the document.
+	// Use query.MustParsePredicate(`{"field": "value"}`) to populate this field.
+	Dependency Predicate
 	// Filterable defines that the field can be used with the `filter` parameter.
 	// When this property is set to `true`, you may want to ensure the backend
 	// database has this field indexed.
```
smyrman: ACK

schema/query/fuzz_test.go
```diff
@@ -0,0 +1,9 @@
+package query
```
smyrman: IF only access to the public interface is needed, switch to `package query_test`.

schema/query/fuzz.go
```diff
@@ -0,0 +1,51 @@
+// +build gofuzz
+
+package query
```
smyrman: IF only access to the public interface is needed, switch to `package query_test`.

schema/query/predicate.go
```diff
@@ -0,0 +1,387 @@
...
+
+// Validate implements Expression interface.
+func (e Regex) Validate(validator schema.Validator) error {
+	return validateValue(e.Field, e.Value, validator)
+}
+
+// String implements Expression interface.
+func (e Regex) String() string {
+	return quoteField(e.Field) + ": {" + opRegex + ": " + valueString(e.Value) + "}"
+}
```
smyrman: ACK

schema/query/predicate_parser.go
```diff
@@ -476,23 +470,23 @@ func (p *parser) expect(c byte) bool {
...
 	}
 	return 0
 }
 
 // eatWhitespaces advance the cursor position pos until non printable characters are met.
-func (p *parser) eatWhitespaces() {
+func (p *predicateParser) eatWhitespaces() {
 	for p.more() {
 		switch p.query[p.pos] {
 		case ' ', '\n', '\r', '\t':
```
smyrman: ACK

schema/query/predicate_parser_test.go
```diff
@@ -344,8 +360,9 @@ func TestParse(t *testing.T) {
 				return
 			}
 			// Parse the result of query.String()
+			t.Logf("%#v", got)
 			str := got.String()
-			got, err = Parse(str)
+			got, err = ParsePredicate(str)
 			if err != nil {
 				t.Errorf("unexpected error for reparsed query `%v`: %v", str, err)
 			}
```
smyrman: ACK

schema/query/predicate_test.go
```diff
@@ -137,15 +137,15 @@ func TestString(t *testing.T) {
...
+		`{"$and": [{"foo": "bar"}, {"foo": "baz"}]}`:              `{$and: [{foo: "bar"}, {foo: "baz"}]}`,
 		`{"foo": "bar", "$or": [{"bar": "baz"}, {"bar": "foo"}]}`: `{foo: "bar", $or: [{bar: "baz"}, {bar: "foo"}]}`,
 		`{"foo": ["bar", "baz"]}`:                                 `{foo: ["bar","baz"]}`,
 	}
 	for query, want := range tests {
-		q, err := Parse(query)
+		q, err := ParsePredicate(query)
 		if err != nil {
 			t.Errorf("Unexpected error for query `%v`: %v", query, err)
 		}
```
smyrman: ACK

schema/query/predicate_validator_test.go
```diff
@@ -93,7 +93,7 @@ func TestValidateErrors(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
-		q, err := Parse(tt.query)
+		q, err := ParsePredicate(tt.query)
 		if err != nil {
 			t.Errorf("Unexpected parse error for `%v`: %v", tt.query, err)
 			continue
```
smyrman: ACK

schema/query/projection.go
```diff
@@ -0,0 +1,37 @@
+package query
+
+import (
+	"github.com/rs/rest-layer/schema"
+)
+
+type Projection []ProjectionField
+
+type ProjectionField struct {
```
smyrman: Missing doc-string on public types (improvement for later?)

schema/query/projection.go
```diff
@@ -0,0 +1,37 @@
...
+		if err := pf.Validate(validator); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// String output the projection in its DSL form.
+func (Projection) String() string {
+	return "" // XXX
```
smyrman: Does this need to be fixed, or documented as not being implemented in the doc-string?

schema/query/projection_evaluator.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+	if s, ok := def.Validator.(schema.FieldSerializer); ok {
+		val, err = s.Serialize(val)
+		if err != nil {
+			return nil, fmt.Errorf("%s: %v", pf.Name, err)
+		}
+	}
+	return val, nil
+}
```
smyrman: ACK

schema/query/projection_evaluator_test.go
```diff
@@ -0,0 +1,205 @@
...
+			if err != nil {
+				return
+			}
+			got, _ := json.Marshal(payload)
+			if string(got) != tc.want {
+				t.Errorf("Eval:\ngot:  %v\nwant: %v", string(got), tc.want)
+			}
+		})
+	}
+}
```
smyrman: ACK

schema/query/projection_parser.go
```diff
@@ -0,0 +1,329 @@
...
+func (p *projectionParser) eatWhitespaces() {
+	for p.more() {
+		switch p.exp[p.pos] {
+		case ' ', '\n', '\r', '\t':
+			p.pos++
+			continue
+		}
+		break
+	}
+}
```
smyrman: ACK

schema/query/projection_parser_test.go
```diff
@@ -0,0 +1,219 @@
...
+			}
+			if err != nil {
+				return
+			}
+			if !reflect.DeepEqual(pr, tc.want) {
+				t.Errorf("Projection:\ngot:  %#v\nwant: %#v", pr, tc.want)
+			}
+		})
+	}
+}
```
smyrman: ACK

schema/query/projection_validator_test.go
```diff
@@ -0,0 +1,54 @@
...
+			pr, err := ParseProjection(tc.projection)
+			if err != nil {
+				t.Errorf("ParseProjection unexpected error: %v", err)
+			}
+			if err = pr.Validate(s); !reflect.DeepEqual(err, tc.err) {
+				t.Errorf("Projection.Validate error = %v, wanted: %v", err, tc.err)
+			}
+		})
+	}
+}
```
smyrman: ACK

schema/query/query.go
smyrman: Can we drop New for now?

schema/query/resolver.go
```diff
@@ -0,0 +1,137 @@
...
+				return err
+			}
+		}
+		payloadsWrapper[0] = p
+		if err := handler(payloadsWrapper, validator); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
smyrman: ACK

schema/query/sort.go
```diff
@@ -0,0 +1,66 @@
...
+		field := validator.GetField(sf.Name)
+		if field == nil {
+			return fmt.Errorf("invalid sort field: %s", sf.Name)
+		}
+		if !field.Sortable {
+			return fmt.Errorf("field is not sortable: %s", sf.Name)
+		}
+	}
+	return nil
+}
```
smyrman: ACK

schema/query/sort_test.go
```diff
@@ -0,0 +1,78 @@
...
+			sort, err := ParseSort(tt.sort)
+			if err != nil {
+				t.Errorf("unexpected parse error: %v", err)
+			}
+			if err := sort.Validate(s); !reflect.DeepEqual(err, tt.err) {
+				t.Errorf("unexpected validate error:\ngot:  %#v\nwant: %#v", err, tt.err)
+			}
+		})
+	}
+}
```
smyrman: ACK

schema/query/utils_test.go
```diff
@@ -2,36 +2,35 @@ package query
...
+	for i := range cases {
+		tc := cases[i]
+		t.Run(tc.name, func(t *testing.T) {
+			_, ok := isNumber(tc.input)
+			if ok != tc.want {
+				t.Errorf("isNumber = %v, wanted %v", ok, tc.want)
+			}
+		})
+	}
 }
```
smyrman: ACK

schema/query/window.go
```diff
@@ -0,0 +1,28 @@
+package query
+
+type Window struct {
```
smyrman: Missing doc-string for Window (improvement for later?)

schema/query/window.go
```diff
@@ -0,0 +1,28 @@
...
+		page = 1
+	}
+	if skip < 0 {
+		skip = 0
+	}
+	return &Window{
+		Offset: (page-1)*perPage + skip,
+		Limit:  perPage,
+	}
+}
```
smyrman: ACK

vendor/github.com/rs/rest-layer-hystrix/hystrix.go
```diff
@@ -0,0 +1,148 @@
+// Package restrix is a REST Layer resource storage wrapper to add hystrix support
```
smyrman: Is the vendor files added by mistake?

rs: I vendored those two small libs to make tests passes and avoid sync issues with those two projects. The rest-layer-mem is actually only useful in the context of unit tests and examples, so it should not be an issue. For hystrix, we might just move the example in the rest-layer-hystrix package instead, and unvendor it.

examples/auth-jwt/main.go
```diff
@@ -181,16 +179,14 @@ func (a AuthResourceHook) OnDelete(ctx context.Context, item *resource.Item) err
...
 		return resource.ErrUnauthorized
 	}
-	// Add a lookup condition to restrict to impact of the clear on objects owned by this user
-	lookup.AddQuery(query.Query{
-		query.Equal{Field: a.UserField, Value: user.ID},
-	})
+	// Add a predicate to the query to restrict to impact of the clear on objects owned by this user
+	q.Predicate = append(q.Predicate, query.Equal{Field: a.UserField, Value: user.ID})
 	return nil
 }
```
rs: This ack flood make the rest of the review pretty hard to parse :)

vendor/github.com/rs/rest-layer-hystrix/hystrix.go
smyrman: Yeah I saw. What you suggest makes sense, but I don't mind if you want to keep it as is for now.

resource/storage.go
rs: I could make this "must" a "should" and explain why. Some storer (like SQL) might want to ignore that recommendation (i.e.: thru an option) to avoid doing a `SELECT *` with a clear statement that it might break interop with some intermediate storers.

schema/query/fuzz.go
rs: What would be the benefit here?

schema/query/query.go
rs: I'm using it in unit test of the storers.

smyrman: Alright

schema/query/fuzz.go
smyrman: If you look at tests as documentation, it makes names read as they would when you use the package elsewhere. E.g. you would import the`query` package and use `query.And` over `And` etc., which in some cases can make the names more useful. It also forces you to stick with the public API, which is usually what you want to keep working as is when you do changes.

smyrman: It's a common pattern used throughout the Go standard library.

rs: I totally agree with that, for tests. This one is not a test but a build flag for go-fuzz. I'm not even sure it would work actually.

smyrman: Yeah, I must have missed the filename during review. Sorry about that! I have no experience with go-fuzz.

smyrman:
As a FYI it's safe to click the "Delete branch", as the branch may later be restored by clicking "restore branch" on the PR. It makes the branch-drop-down on the "Code" tab less cluttered.

PR #119, opened 2017-08-03
Fix demo

### LINKED ISSUE #118: Oops... Documentation example fails
$ go run examples/demo/main.go

{"time":"2017-08-03T07:48:19+03:00","level":"fatal","message":"Invalid API configuration: users.posts: schema compilation error: meta.published: unknown query field"}
exit status 1

Please, fix it!

rs:
The demo uses a dependency referencing a field from a parent schema.
This does not work with the current implementation, and I don't recall
it working in the past either. We might want to make such dependency
work, but it might require some major changes.

I thus removed the sub-schema and made the demo more realistic.

Fix #118

### DIFF
--- a/examples/demo/main.go
+++ b/examples/demo/main.go
@@ -4,6 +4,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"net/url"
 	"time"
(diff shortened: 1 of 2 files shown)


examples/demo/main.go
```diff
@@ -26,19 +27,12 @@ var (
...
-				// This is a field hook called when a new user is created.
-				// The schema.NewID hook is a provided hook to generate a
-				// unique id when no value is provided.
-				OnInit: schema.NewID,
 				// The Filterable and Sortable allows usage of filter and sort
 				// on this field in requests.
 				Filterable: true,
 				Sortable:   true,
 				Validator: &schema.String{
-					Regexp: "^[0-9a-v]{20}$",
```
smyrman: I was curious to why it was a-v...

rs: base32

Issue #120, opened 2017-08-09
Add ErrForbiden to resource package


smyrman:
It wold be good to add an `ErrForbidden` to the `resource` package so that it may be returned by resource hooks and understood by the `rest` (and `graphql`) package when generating a response.

To avoid having to state a case for _every_ possible HTTP code though, maybe it would be better to let the error be an integer in the `resource` package?
```go
type Error int

func (status Error) Error() string {
        return http.StatusText(status)
}

var (
        ...
        ErrForbidden Error = Error(http.StatusForbidden)
        ...
)
```
Then we don't need a switch-case error-handler in the `rest` package for this specific issue, as we could just convert it like this:
```go
if status, ok := err.(resource.Error); ok {
        return Error{int(status), http.StatusText(int(status)), nil}
} else if err != nil {
        return &Error{520, err.Error(), nil}
} else {
        return nil
}

```

It could also make sense to remove `ErrUnauthorized`, as the handlers for it (in the `rest` package at least) do not set the required `WWW-Authenticate` header. Nor does rest-layer have any built-in knowledge for what the value of this header should be.

We would need to update some examples to reflect this:

I think it should probably be up to the authentication layer (and the authentication layer only) to return a 401 response, and if a request goes past the authentication layer, even if by mistake, and hits a resource hook that e.g. need access to the current user, we return a 403 forbidden.

rs:
Agreed with all the points.

I did not use the HTTP status codes in the `resource.Error` type because I didn't want to "limit" or tie this package to the HTTP transport. I'm find revisiting that.

smyrman:
Yeah I took it that was the reason. We can maybe start out with removing `ErrUnauthorized` (since we can't handle it 100% correct) and add `ErrForbidden` in it's place. Then we can consider _later_ if there is a desire to _optimise_ for HTTP error handling and make that _easier_ than non-http error handling.

PR #121, opened 2017-08-13
Replaces ErrUnauthorized by ErrForbidden (breaking change)

smyrman:
This is just a start on #120.

In the resource and rest packages, ErrUnauthorized is replaced by
ErrForbidden. It is presumably more useful to let rest-layer hooks
return the new error over an error that will be translated into a 401
response by the `rest` package. Despite it's name, 401 Unauthorized
responses should only be return when a user is not authenticated. 403
Forbidden errors is a much better fit for any kind of permission (or
authentication issues) that is discovered as late as within a resource
hook.

401 Unauthorized HTTP response, is best returned by an authentication
middleware only, as this middleware will have the opportunity to set an
appropriate WWW-Authenticate challenge for the given application, while
the rest-layer `rest` package may not.

~PS! This change does not update the examples!~

### DIFF
--- a/examples/auth-jwt/main.go
+++ b/examples/auth-jwt/main.go
@@ -66,7 +66,7 @@ func NewJWTHandler(users *resource.Resource, jwtKeyFunc jwt.Keyfunc) func(next h
 			// Lookup the user by its id
 			ctx := r.Context()
 			user, err := users.Get(ctx, userID)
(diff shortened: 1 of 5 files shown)


smyrman:
Yeah, tests and examples both need updating before this can go in.

smyrman:
@rs, this has now been updated.

Issue #125, opened 2017-09-05
Add Decimal/Numeric schema type


Dragomir-Ivanov:
Hi There,
It would be good if Rest-Layer supports Decimal/Numeric schema type, so it can be used in financial applications. DBs supporting decimals naively:
```
MongoDB 3.4+ - NumberDecimal
SQLite3 - NUMERIC
ElasticSearch - scaled_float
```
Any comments appreciated.

smyrman:
It might be best to do this either via the application itself, or via the different storage backend libraries.

E.g. you could add a type `NumericDecimal` that implements the `schema.FieldValidator` and `schema.FieldSerializer` interfaces to `rs/rest-layer-mongo`.

smyrman:
PS!  I would add it to an application first, and if it works well, contribute it back to one of the storage backends (and potentially get it cleaned up a bit).

Dragomir-Ivanov:
Thanks for the comments @smyrman. Will investigate and try to add it in the application. I understand that `Serialize` function is called before data is passed from the `Storer` to the REST response, however I don't understand how things are serialized from other way around. E.g. from REST request to `Storer`. My idea is: 
1. Numeric data is passed as string in JSON request. 
2. Rest-Layer sees that field is `Schema.Numeric` and then validates and converts it to this type.
3. Mongo backend converts `Schema.Numeric` to Decimal128 mongo type.

Thanks for the help, still new to Rest-Layer and Go but is seems great. Is Rest-Layer used somewhere in production?

smyrman:
> E.g. from REST request to Storer. My idea is:

That is the Validator

smyrman:
> I understand that Serialize function is called before data is passed from the Storer 

Correct:-) So here you could return e.g. a `json.RawMessage` to make it simple, or any type that implements the `json.Marshaller` interface.

Dragomir-Ivanov:
Thanks,
Just figured out this myself. This is my first naive implementation:
```
// Decimal type, works with Mongo 3.4+ for now only
type Decimal struct {
	StoreBinary bool
}

// Validate implements FieldValidator
func (v Decimal) Validate(value interface{}) (interface{}, error) {
	s, ok := value.(string)
	if !ok {
		return nil, errors.New("invalid type")
	}
	ip, err := bson.ParseDecimal128(s)
	if err != nil {
		return nil, errors.New("invalid Decimal format")
	}

	return ip, nil
}

// Serialize implements FieldSerializer.
[...]

PR #128, opened 2017-09-07
json: Unmarshall JSON numbers as json.Number type

Dragomir-Ivanov:
This could theoretically allow us to support other number types, such as `Decimal` and being able to transport it via JSON:
```
{
    "value" : 1.23
}
```

instead quoted number
```
{
    "value" : "1.23"
}
```

@smyrman  Still not sure about this.

### DIFF
--- a/rest/routing.go
+++ b/rest/routing.go
@@ -189,6 +189,10 @@ func (r *RouteMatch) Query() (*query.Query, *Error) {
 			q.Predicate = append(q.Predicate, query.Equal{Field: rp.Field, Value: rp.Value})
 		}
 	}
+	rsc := r.Resource()
+	if rsc == nil {
+		return nil, &Error{500, "missing resource", nil}
+	}
 	// Parse query string params.
 	switch r.Method {
 	case "HEAD":
@@ -201,7 +205,7 @@ func (r *RouteMatch) Query() (*query.Query, *Error) {
 				return nil, &Error{422, err.Error(), nil}
 			}
 			limit = l
-		} else if l := r.Resource().Conf().PaginationDefaultLimit; l > 0 {
+		} else if l := rsc.Conf().PaginationDefaultLimit; l > 0 {
 			limit = l
 		}
 		skip := 0
@@ -226,7 +230,7 @@ func (r *RouteMatch) Query() (*query.Query, *Error) {
 	if sort := r.Params.Get("sort"); sort != "" {
 		s, err := query.ParseSort(sort)
 		if err == nil {
-			err = s.Validate(r.Resource().Validator())
+			err = s.Validate(rsc.Validator())
 		}
 		if err != nil {
 			return nil, &Error{422, fmt.Sprintf("Invalid `sort` parameter: %s", err), nil}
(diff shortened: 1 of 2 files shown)


Dragomir-Ivanov:
Seems like I missed `query/` tests.

smyrman:
No the intend was to start a discussion on it first.

smyrman:
But I suppose the PR makes the discussion easy to have.

smyrman:
I think it gives nice leeway, but I want to hear @rs opinion. Is it worth switching to `json.Number` to allow more leeway in e.g. fixed decimal implementations for applications and/or Storere backends? The implementations could of-course still _choose_ to rely on strings.

rs:
Wouldn't it be better to keep current integer/float validation untouched and deal with decimals using a specific validator that would enforce the value to be un/serialized as a string to keep the right representation?

Dragomir-Ivanov:
@rs Agreed, it is sane to keep "custom" types always encoded as strings. Will make a PR to rs-layer-mongo soon.

smyrman:
First of all, I think this does not need to be done _now_. You can perfectly well implement a Decimal type that supports only string encoding, and that would probably be the most useful, and the best place to start, as you have already pointed out @Dragomir-Ivanov.

My intent with reopening #125  rest-layer side, was just so we could discuss if it was also worthwile to consider adding leeway in rest-layer to do more than that (in good time). Thanks for this code demonstration though.

> deal with decimals using a specific validator that would enforce the value to be un/serialized as a string to keep the right representation?

@rs , how could you do that when the payload is  always decoded into a `map[string]interfave{}`? The only way to do parse a numeric JSON tag into a number specifically for one field would be to use structs as schemas, or massivly complicate the decoding process.

Therefore _if_ we want a fixed Decimal type that supports JSON numeric encoding as well as string encoding, and maybe we don't btw, then this is probably the best aproach as it leaves all numbers in their most pure original string form until they reach the validator.

rs:
If decimals are stored as strings there is nothing special JSONwize to have them parsed. It is the role of the validator to have them checked and eventually transformed if necessary.

smyrman:
@rs, yes. That has always been clear.

smyrman:
Or depends what you mean as "stored as string" @rs. `json.Number` is a string in go, just as the whole JSON payload is a string, or byte slice in go.

The point is that if we want to parse  number literals in JSON (a string or byte slice withot surrounding double quotes), today that number is parsed directly into a float64 or int64 before it even reach the validator, is it not?

If we want to add leeway to add numeric FieldValidators with a higher or different precision than float64, we need to use `json.Number` when parcing into `map[string]interface{}` to avoid loosing precision.

Fixed decimal is only an example. Another example would be big ints/ floats.

smyrman:
And btw, maybe the answer is we don't need such leeway. Maybe such numeric types would be restricted to use string literals in JSON when combined with rest-layer, meaning a byte slice where the first and last char is a double quote.

rs:
I would keep it simple. I see no practical use-case where the current behavior would be an issue. Do you have some?

smyrman:
Not an issue as such, as this is just a discussion on style.

Support for big int/float could matter for scientific applications, and fixed decimal could be essential for financial application, but there is no case that _demands_ these types to be encoded as a numeric literal in JSON.

Let's close it for now anyway :+1:

Issue #131, opened 2017-09-08
Content-Location not generated correctly with mongo IDs.


rs:
The `Content-Location` header is not serialized correctly when the ID is a mongo ID.

```
$ bat POST :2015/api/as Authorization:"Bearer ${KEY}" name=n1  
POST /api/as HTTP/1.1
Host: localhost:2015
Accept: application/json
Accept-Encoding: gzip, deflate
Authorization: Bearer XXX
Content-Type: application/json
User-Agent: bat/0.1.0

{"name":"n1"}

HTTP/1.1 201 Created
Last-Modified: Thu, 07 Sep 2017 20:28:37 GMT
Server: Caddy
Content-Length: 45
Content-Location: /as/ObjectIdHex("59b1abf5cee0fb0001ea732e")
Content-Type: application/json
Date: Thu, 07 Sep 2017 20:28:37 GMT
Etag: "5efa3a3f0f355cef35d7ca2464d4836d"

{
  "id": "59b1abf5cee0fb0001ea732e",
  "name": "n1"
}
```

Dragomir-Ivanov:
The problem lies in `encoding/bson` Go package (bug maybe?):
```
// String returns a hex string representation of the id.
// Example: ObjectIdHex("4d88e15b60f486e428412dc9").
func (id ObjectId) String() string {
	return fmt.Sprintf(`ObjectIdHex("%x")`, string(id))
}
```

rs:
We might need to wrap it then.

Dragomir-Ivanov:
@rs Interestingly how `bson.ObjectId` leaks into rest-layer. Shouldn't all such occurrences be wrapped within `github.com/rs/rest-layer-mongo` `ObjectID` type?

rs:
I think so.

Dragomir-Ivanov:
@rs Thanks, will try to fix this issue myself.

smyrman:
Not sure it's in the mongo package.  I used the  rest-layer-mongo.ObjectIDField. Should it not call the serializer method in rest-layer on the location URLgeneration?

https://github.com/rs/rest-layer-mongo/blob/f301bb23e3cb499e6ddd0c9c7210a3d158570f9f/id.go#L57

This method does return a string:
https://godoc.org/gopkg.in/mgo.v2/bson#ObjectId.Hex

Dragomir-Ivanov:
@smyrman Please take a look at the open issue in `rest-layer-mongo`

rs:
[...]

PR #132, opened 2017-09-11
Fix wrong `Content-Location` with MongoDB  (#131)

Dragomir-Ivanov:


rest/method_post.go
```diff
@@ -47,6 +48,14 @@ func listPost(ctx context.Context, r *http.Request, route *RouteMatch) (status i
 	}
 	// See https://www.subbu.org/blog/2008/10/location-vs-content-location
 	headers = http.Header{}
-	headers.Set("Content-Location", fmt.Sprintf("%s/%s", r.URL.Path, item.ID))
+	itemID := item.ID
+	if f := rsrc.Validator().GetField("id"); f != nil {
+		if s, serialize := f.Validator.(schema.FieldSerializer); serialize {
```
smyrman: I think the following might be more idiomatic:

`s, ok := f..Validator.(schema.FieldSerializer); ok`

However, this is also OK.

rest/method_post.go
```diff
@@ -47,6 +48,14 @@ func listPost(ctx context.Context, r *http.Request, route *RouteMatch) (status i
...
+			if tmp, err := s.Serialize(itemID); err == nil {
+				itemID = tmp
+			}
```
smyrman: Do we need to log this error? @rs, how is this handled elsewhere?

smyrman:
Also, please squash the commits before we do the final merge.

`git fetch && git rebase -i origin/master`. Mark all commits except the first one as "fixup" (or "squash" to merge & modify the commit message). Force-push the new version of your branch.

Dragomir-Ivanov:
@smyrman Your git command didn't worked. Will make new pull request for now.

smyrman:
You can squash and merge, but the commit message doesn't get as pretty, as you can't edit it. When dealing with Git history, there are really two must-learn commands, and that is:
- `git rebase -i <remote>/<branch>` interactive rebase
- `git commit --amend`, change the last commit :-)

https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

smyrman:
Please keep the same PR, it's easier for review! to force-push (the current branch), use `git push -f`. The more explicit vertsion is `git push -f <remote> <branch>`

Dragomir-Ivanov:
Should I make changes in topic branch, or keeping them in `master`?
From what I have read about Git, it is wrong to change already published commits.
When I tried `git rebase -i origin/master` it showed only the last commit, and not the whole history where I could pick the commits I want to squash. 
Anyway, I am ignorant with Git for now, so any help appreciated. Thanks :)

smyrman:
Hi. I think it's a bit individual, and there are different work flows. @rs work-flow for instance, is probably slightly different than mine.

> When I tried git rebase -i origin/master it showed only the last commit, and not the whole history.

It also depends on what "origin" is in your context (`git remote -v` will show this).  Basically the argument is just which commit you want to rebase on top of. E.g. if you write HEAD~5, you will see, and be able to rebase the last 5 commits, or if you write some other SHA-1, it will ttry to rebase on top of that commit. If you do many concurrent PRs to rest-layer, you also need to use one branch per PR, rather than your master branch.

Usually though, you would want to rebase on top of the branch that you are going to push to. So if you named the remote `git@github:rs/rest-layer.git` as `foo` (see `git remote -v` to list your remotes), you would want to rebase on top of `foo/master`, and solve any potential conflicts.

The `-i` is only needed if you want to edit the order, wording, merge or even remove commits. This is useful if you have more than one commit different from master.

> From what I have read about Git, it is wrong to change already published commits.

Yes, but it depends on what you define as "published". Reached a branch? Passed review? Merged to master?

I define "published" as passed review & merged to a "stable" branch, e.g. master. By changing the history, you get to reword the commit messages, and let each commit be a single "path" that does one thing, and one thing only, to the code base. E.g. adds a complete feature, or fixes one bug. The Linux kernel is developed in a similar way, although the way they review code is very different; they use email and formatted patches!

For a bit of history, when Git was new, it was very easy to shoot yourself in the foot by changing history and force push, so you didn't want to do that. Now you have tools to prevent you doing these mistakes as you can, via e.g. GitHub, protect branches from history changes (typically the master branch), and enforce CodeReview and/or CI checks.

Dragomir-Ivanov:
Thanks for the extensive explanation @smyrman . I guess I took your commands too ultimate and didn't do `git rebase -i origin/master HEAD~3` 

>  PS! Always run git fetch foo first, to ensure you have the latest history from foo available localy!
Hmm, since this is my fork, doesn't I have already latest history available locally? Do I make something wrong? What I do, is I make a clone of say `rest-layer` in my Github account so I have only these 2 default remotes:
```
drago@T430s:~/temp/rest-layer$ git remote -v
origin	https://github.com/Dragomir-Ivanov/rest-layer.git (fetch)
origin	https://github.com/Dragomir-Ivanov/rest-layer.git (push)
```
Then I push all my work onto these branches, and when satisfied make a pull request from the Github GUI.
I will need so practice to iron things out. Thanks again man.

smyrman:
Ah... I mean fetch the latest version of the upstream `rs/rest-layer` repo in case there are any changes done there. E.g. you could do this:
```sh
$ git remote add rs https://github.com/rs/rest-layer.git
$ git fetch rs
```

As a FYI, my workflow (assuming the remote is now names `rs`) for working on a new feature would probably be something like:
```sh
$ git checkout -b my-new-feature
$ git fetch rs
$ git reset --hard rs/master   # Careful! drops all local changes/history, and checks out the up-stream master at rs.
... write my code ...
$ git commit -a
$ git push --set-upstream origin my-new-feature
```
Sometimes in a different order.

After that, `git add`, `git commit`, `git commit --ammend`, `git rebase -i rs/master` and `git push -f` would be my preferred commands for working towards a final patch (or set of patches). You might prefer a different work-flow yourself, or you might prefer using a graphical tool.

Issue #129, opened 2017-09-07
Embedding of connected resource is broken


smyrman:
Using the `rest` package, given a resource "as" with a sub-resource "bs", when fetching "as" with `fields=id,bs`, _all_ "bs" are included in _all_ entries. Instead, I wold have expected `bs` to be filtered by the `id` of the associated `a`.

# Test
using bat which is a Go clone of HTTPie:

```yaml
----------------------------------------------------------------------
$ bat POST :2015/api/as Authorization:"Bearer ${KEY}" name=n1  
POST /api/as HTTP/1.1
Host: localhost:2015
Accept: application/json
Accept-Encoding: gzip, deflate
Authorization: Bearer XXX
Content-Type: application/json
User-Agent: bat/0.1.0

{"name":"n1"}

HTTP/1.1 201 Created
Last-Modified : Thu, 07 Sep 2017 20:28:37 GMT
Server : Caddy
Content-Length : 45
Content-Location : /as/ObjectIdHex("59b1abf5cee0fb0001ea732e")
Content-Type : application/json
Date : Thu, 07 Sep 2017 20:28:37 GMT
Etag : "5efa3a3f0f355cef35d7ca2464d4836d"

{
  "id": "59b1abf5cee0fb0001ea732e",
  "name": "n1"
}

----------------------------------------------------------------------
$ bat POST :2015/api/as Authorization:"Bearer ${KEY}" name=n2
POST /api/as HTTP/1.1
Host: localhost:2015
Accept: application/json
Accept-Encoding: gzip, deflate
Authorization: Bearer XXX
Content-Type: application/json
User-Agent: bat/0.1.0

{"name":"n2"}

HTTP/1.1 201 Created
Etag : "9e969caf8cabb3303886ca7fce114668"
Last-Modified : Thu, 07 Sep 2017 20:28:40 GMT
Server : Caddy
Content-Length : 45
Content-Location : /as/ObjectIdHex("59b1abf8cee0fb0001ea732f")
Content-Type : application/json
Date : Thu, 07 Sep 2017 20:28:40 GMT


{
  "id": "59b1abf8cee0fb0001ea732f",
  "name": "n2"
}

----------------------------------------------------------------------
$ bat POST :2015/api/as/59b1abf5cee0fb0001ea732e/bs Authorization:"Bearer ${KEY}" name=b1
POST /api/as/59b1abf5cee0fb0001ea732e/bs HTTP/1.1
Host: localhost:2015
Accept: application/json
[...]

PR #141, opened 2017-09-23
Fix issue 129 by adding Field to schema.Connection

smyrman:
Fix issue #129.

### DIFF
--- a/schema/query/projection_evaluator_test.go
+++ b/schema/query/projection_evaluator_test.go
@@ -55,6 +55,7 @@ func (r resource) Validator() schema.Validator {
 func TestProjectionEval(t *testing.T) {
 	r := resource{
 		validator: schema.Schema{Fields: schema.Fields{
+			"id":     {},
 			"simple": {},
 			"parent": {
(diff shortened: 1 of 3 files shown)


schema/query/projection_evaluator_test.go
```diff
@@ -165,11 +169,18 @@ func TestProjectionEval(t *testing.T) {
...
+			`connection{name}`,
+			`{"id":"a","simple":"foo"}`,
+			nil,
+			`{"connection":[{"name":"second"},{"name":"forth"}]}`,
+		},
+		{
+			"Connection#2",
 			`connection{name}`,
-			`{"simple":"foo"}`,
+			`{"id":"b","simple":"foo"}`,
```
smyrman: We could maybe add a test to test filters / embedding together.

smyrman:
This PR seams to somehow be missing the changes I had done to the `resource` package 😞 That is where The Field get's set on Bind.

The unit-tests still work because they don't rely on the resource Bind method. I will have to do another PR to fix it.

Issue #140, opened 2017-09-22
Adding a bool parameter to a field does not parse correctly


smyrman:
It seams the current query parser does not handle bool values in field parameters.

Example:
 
If I use a boolean validator like this:
```go
schema.Schema{
	Fields: schema.Fields{
		"items": {
			 Params := schema.Params{
				"expand": {Validator: &schema.Bool{}},
			},
		},
	},
}
```

If I now ask for `fields=items(expand:true)`, I will get an error when trying to filter on a `rest` resource:
```
HTTP/1.1 422 Unprocessable Entity
Content-Length : 81
Content-Type : application/json
Date : Fri, 22 Sep 2017 07:04:44 GMT
Server : Caddy
Vary : Origin


{
  "code": 422,
  "message": "Invalid `fields` parameter: looking for value at char 13"
}
```

smyrman:
As a FYI, I have worked around this using an integer that may be 0 or 1, but we should probably address this? I think this also relates to #139, or at least, they are both issues with the current query parser implementation.

smyrman:
Needs case for `t` and `f` here:

https://github.com/rs/rest-layer/blob/b0185e617e9718fea09957f10126b020d04bc033/schema/query/projection_parser.go#L215

rs:
Good catch, will you fix it or should I?

smyrman:
If you want to fix it, that would be great. If not, I _might_ have some time next week.

There is a need for a  new method `projectionParser.parseBool()`, although I see you got an implementation of this for another parser type here:

https://github.com/rs/rest-layer/blob/b0185e617e9718fea09957f10126b020d04bc033/schema/query/predicate_parser.go#L378

PR #146, opened 2017-10-03
Add support for bool parameters

smyrman:
This commit adds support for adding bool parameters to a schema field.
For consistency with other FieldValidators, the error message was also
changed to state "boolean" as lower case.

Fixes issue #140.

### DIFF
--- a/schema/query/projection.go
+++ b/schema/query/projection.go
@@ -75,6 +75,8 @@ func (pf ProjectionField) String() string {
 				buf.WriteString(strconv.Quote(v))
 			case float64:
 				buf.WriteString(strconv.FormatFloat(v, 'f', -1, 64))
+			case bool:
+				buf.WriteString(fmt.Sprintf("%t", v))
 			default:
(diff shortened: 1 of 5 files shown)


schema/alloff_test.go
```diff
@@ -35,7 +35,7 @@ func TestAllOfValidator(t *testing.T) {
 			Name:      `{Bool, String}.Validate("")`,
 			Validator: schema.AllOf{&schema.Bool{}, &schema.String{}},
 			Input:     "",
-			Error:     "not a Boolean",
+			Error:     "not a boolean",
```
rs: Boolean is a last name and does always take a cap.

smyrman: Right, from George Boole it seams. Given the Oxford dictionary also agrees with you I suppose I will change it back 👍

smyrman:
Reverted Boolean -> boolean rename, and added one more test case in `projection_evaluator_test.go`

Issue #139, opened 2017-09-20
Filtering on fields of type dictionary does not work


torie:
I am trying to filter responses by a specific key-value combination for a dictionary field, which seems to not be working. 

## Steps to reproduce:

Define a resource with a schema.Field with the following specification:

```golang
"foobar": {
        Description: "A set of key values.",
        Validator: &schema.Dict{
	        KeysValidator:   &schema.String{MinLen: 1},
		ValuesValidator: &schema.String{MinLen: 1},
	},
	Filterable: true,
},
```

Make a request to get resources which has a specific key-value pair in the `foobar` dictionary.

```http
http://myapi.com/myresource?filter={"foobar":{"foo": "bar"}}
```

The following error occurs.

```
"Invalid `filter` parameter: foobar: invalid query expression: not a dict"
```

---

I've done some digging, and the error occurs in schema/dict.go when the `func (v Dict) Validate(...)`  function attempts to check if the incoming value is of type map[string]interface{}. At this point the value has been parsed into map[string]query.Value which fails the test.

```golang
// Validate implements FieldValidator interface.
func (v Dict) Validate(value interface{}) (interface{}, error) {
	dict, ok := value.(map[string]interface{}) // <- value is of type map[string]query.Value here
	if !ok {
		return nil, errors.New("not a dict")
	}
        ...
```

smyrman:
For us (me and @torie at Searis), there are actually two cases we would like to work:

- Case 1:  exact match `http://myapi.com/myresource?filter={"foobar":{"foo": "bar"}}`

This is actually not too important.

- Case 2: search for the value of a key `http://myapi.com/myresource?filter={"foobar.foo": "bar"}`

This is more relevant, as it would allow us to use a dictionary to hold arbitrary tag/label values and filter based on individual fields.

smyrman:
As for case2, the first step would be to make sure the following code can find the field of a `schemaDict` or `schema.Object`:

https://github.com/rs/rest-layer/blob/5b33327a0fb07b058fa7f55e5c7798c1a9b07164/schema/schema.go#L57

Most likely this would involve these two types implementing a `GetField` method and check for it in a similar way we already do if a field have a `Schema`.

smyrman:
Starting on `case 2` only for now.

smyrman:
> "Invalid `filter` parameter: foobar: invalid query expression: not a dict"

[...]

Issue #77, opened 2017-01-02
Question: Field.Schema v.s. Field.Validator = schema.Object


smyrman:
What are the differences between
```go
schema := schema.Schema{
        Fields: {
        }
}
```

rs:
Good question. It was add by @yanfali In #24, he can certainly answer.

yanfali:
Primarily use case. I wanted to use a schema in array types. Previously you could only use a type which didn't support schema.

smyrman:
Ah, ok so `Field.Schema` is the initial support for sub-schemas here. I actually like the `Object` `FieldValidator` better than having `Field.Schema`, as it's both more consistent, and avoids ambiguous meaning like:

```go
 "meta": Field{Validator: schema.String{}, Schema: subSchema}
```

Would it be sensible to deprecate/remove the Schema attribute from Field - perhaps after tagging a release, as suggested in #65? I assume this could also simplify things internally and avoid handling the same case twice several places? For one I realize we don't handle Field.Schema in the `jsonschema` package.

smyrman:
> For #65 I would rather wait for the Go official way to handle that.

OK, for the manifest part, but tagging a release in Git (e.g. v0.1.0), is something the team have already recommended us to do, but for end-users, rest-layer would be a lot safer/easier to use if you could rely on a tagged release. Especially when we do breaking changes.

rs:
Sure we can tag.

smyrman:
@rs - updated the title & description.

rs:
Are you going to submit a PR for this one?

smyrman:
I don't know if I can/want to figure out the implications on reference checking... there are some other easier issues I would rather look at, e.g. JSON schema ones...

smyrman:
At least not know

smyrman:
@rs Looking at the definition of Object:
```
type Object struct {
    Schema *Schema
}
```

Could we not make Schema implement FieldValidator or at least change the definition to:
```
type Object Schema
```

smyrman:
Maybe it's not so easy btw. At least not without introducing _additional_ breaking changes everywhere...

rs:
I'm all for introducing breaking changes now that simplify the API rather than later.

smyrman:
I have done some background thinking on this, and I have a rather radical proposal:

[...]

PR #151, opened 2017-10-05
Support query on dict/object values (breaking change!)

smyrman:
Starting on #139, case 2.

@rs, some feedback on putting `Values Field` over `ValuesValidator FieldValidator` inside `schema.Dict`. The idea would be to to do the same for `schema.Array` in a later PR. This also allows description fields to be set by `jsonschema`, as well as setting a `Filterable` value, and would become very consistent once we start on #77.

 In addition `schema.AllOf` and `schema.AnyOf` should ideally get a GetFields implementation, although this is a harder problem as they would need to return _multiple_ fields.

So far I have tested filtering to work with some simple manual test code:
-  https://gist.github.com/smyrman/cb6ebcb6d9dbb2e269d648251687bd7b

-----

This commit adds support for looking up field definitions inside
schema.Dict and schema.Object instances. To achieve this,
Dict.ValuesValidator has been deprecated in favour of a new Dict.Values
property that holds a full schema.Field.

As a result, it should be possible to filter on schema.Dict and
schema.Object sub-fields.

PS! This commit does not add support for filtering on fields inside
schema.Array, schema.AnyOf or schema.AllOf; this will come later.


schema/dict.go
```diff
@@ -9,8 +9,13 @@ import (
 type Dict struct {
 	// KeysValidator is the validator to apply on dict keys.
 	KeysValidator FieldValidator
-	// ValuesValidator is the validator to apply on dict values.
-	ValuesValidator FieldValidator
+
+	// ValuesValidator is deprecated, use Values instead to specify a Field. If
+	// used, Values will be set to a required, non-filterable filed on Compile.
+	// ValuesValidator FieldValidator
```
smyrman: We have an _optino_ to add backwards compatibility by keeping the ValuesValidator field until we start on #77, if there is interest for it. It would add some complexity.

smyrman: I.e. I can either comment in or remove this code.

schema/dict.go
```diff
@@ -23,8 +28,19 @@ func (v *Dict) Compile(rc ReferenceChecker) (err error) {
...
-	if c, ok := v.ValuesValidator.(Compiler); ok {
+	/*
+		// Populate v.Values from v.ValuesValidator for backwards-compatibility.
+		if val := v.ValuesValidator; val != nil {
+			log.Println("Deprecated: schema.Dict.ValuesValidator is deprecated in favour of schema.Dict.Values")
+			v.Values = schema.Field{
+				Required:  true,
+				Validator: val,
+			}
+		}*/
```
smyrman: This code would also either be commented in or removed.

smyrman:
Good point @Dragomir-Ivanov. There looks to be no documentation in the README that needs updating, but the doc-comment should most sertanly be updated:

> Dict validates array values

That seems pretty wrong 😆

smyrman:
@rs / @Dragomir-Ivanov PR updated with documentation changes, including a new list of breaking changes.

Sorry about the delay, it's been busy.

smyrman:
@rs, would you prefer `Array.ValuesValidator` to be replaced in the same PR for more consistency on `master`, or are you OK with this arriving in a separate PR?

schema/dict.go
```diff
@@ -67,3 +70,13 @@ func (v Dict) Validate(value interface{}) (interface{}, error) {
 	}
 	return dest, nil
 }
+
+// GetField implements the FieldGetter interface.
+func (v Dict) GetField(name string) *Field {
+	if _, err := v.KeysValidator.Validate(name); err != nil {
+		return nil
+	}
+	// TODO: As part of issue #77, replace Dict.ValuesValidator with a
```
smyrman: TODO is out of date (already done)

smyrman: fixed

schema/field.go
```diff
@@ -93,25 +117,12 @@ type FieldSerializer interface {
...
-	}
-	return nil
+// FieldGetter is used to describe a FieldValidator (or Schema) which allow JSON
+// object values.
+type FieldGetter interface {
+	// GetField returns the validator for the field if the given field name is
+	// present in the schema.
+	//
+	// You may reference sub field using dotted notation field.subfield.
+	GetField(name string) *Field
```
smyrman: `QueryField` could be a better name. At least when it comes to e.g. `Array` and `AnyOf`, we are interested in a Field that can be used to validate query values, not necessarily a field that can validate values.

- https://docs.mongodb.com/manual/tutorial/query-arrays/

smyrman: Requires some rewrite of the whole logic to support better array querying, so we could wait with the rename this until we do the array part, if we need to slightly redo the logic anyway...

Issue #127, opened 2017-09-07
Clear does not support filters or limit


smyrman:
According to the README, a DELETE request on the resource list view has this effect:

> Delete all items from the collection matching the context and/or filters.

However, when I tried locally, it deletes _everything_ and ignores the filter and/or limit.

smyrman:
I tried with the rest-layer-mongo storage backend btw.

smyrman:
It should also be noted that sub-resources that link to a parent object, are not deleted when deleting the parent object. E.g. in case of the #129 schema, 

```
$ bat DELETE :2015/api/as
```
will not delete any of the "bs".

rs:
There is no support for cascading deletion if this is what you are referring to.

The delete with filter bug definitely need to be fixed.

Dragomir-Ivanov:
@smyrman For the cascading deletion case, this is stated on the Rest-Layer documentation site as:  
□ "Cascading deletes on sub resources". However I hope this is still on the radar.

rs:
Yes, it is not yet implemented.

smyrman:
It looks like at least part of the problem is here (maybe all of it):

https://github.com/rs/rest-layer/blob/0df312935ab2631d1813920adf9bb04b328dd071/rest/routing.go#L201

As you can see, there is a case for GET and HEAD, but no case for DELETE. The DELETE case should handle `limit`, `page`, `offset` and `filter`, but ignore fields. In order to reuse as much code as possible, _one_ solution could be to handle the field selectioon bit first, and then use a fall-through:

```go
	case "GET":
		if fields := r.Params.Get("fields"); fields != "" {
		p, err := query.ParseProjection(fields)
		if err == nil {
			err = p.Validate(rsc.Validator())
		}
		if err != nil {
			return nil, &Error{422, fmt.Sprintf("Invalid `fields` parameter: %s", err), nil}
		}
		q.Projection = p
		fallthrough
	case "DELETE":
		limit := -1
		if l, found, err := getUintParam(r.Params, "limit"); found {
			if err != nil {
				return nil, &Error{422, err.Error(), nil}
			}
			limit = l
		} else if l := rsc.Conf().PaginationDefaultLimit; l > 0 {
			limit = l
		}
		// ...
```

Of-course, splitting out in some re-usable functions (or methods) with some built-in error handling could work equally well.

smyrman:
And because I never remember how a `fallthrough` works, here is a reminder:
[...]

PR #152, opened 2017-10-08
Add HTTP HEAD specialization for `resource.ItemList`

Dragomir-Ivanov:
This is generic optimization for HTTP HEAD, where resource list is being queried.
No data is pulled out from the storage layer.

### DIFF
--- a/resource/resource.go
+++ b/resource/resource.go
@@ -342,6 +342,25 @@ func (r *Resource) find(ctx context.Context, q *query.Query, forceTotal bool) (l
 	return
 }
 
+// Count implements Storer interface.
+func (r *Resource) Count(ctx context.Context, q *query.Query) (n int, err error) {
+	if LoggerLevel <= LogLevelDebug && Logger != nil {
+		defer func(t time.Time) {
+			found := n
+			Logger(ctx, LogLevelDebug, fmt.Sprintf("%s.Count(...)", r.path), map[string]interface{}{
+				"duration": time.Since(t),
+				"found":    found,
+				"error":    err,
+			})
+		}(time.Now())
+	}
+	if err = r.hooks.onFind(ctx, q); err == nil {
+		n, err = r.storage.Count(ctx, &query.Query{Predicate: q.Predicate})
+	}
+	r.hooks.onFound(ctx, q, nil, &err)
+	return
+}
+
 // Insert implements Storer interface.
 func (r *Resource) Insert(ctx context.Context, items []*Item) (err error) {
 	if LoggerLevel <= LogLevelDebug && Logger != nil {
--- a/rest/method_head.go
+++ b/rest/method_head.go
@@ -1,0 +1,54 @@
+package rest
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/rs/rest-layer/resource"
+)
+
+// listHead handles HEAD resquests on a resource URL. Returns no payload
+func listHead(ctx context.Context, r *http.Request, route *RouteMatch) (status int, headers http.Header, body interface{}) {
+	forceTotal := false
+	rsc := route.Resource()
+	switch rsc.Conf().ForceTotal {
+	case resource.TotalOptIn:
+		forceTotal = route.Params.Get("total") == "1"
+	case resource.TotalAlways:
+		forceTotal = true
+	case resource.TotalDenied:
+		if route.Params.Get("total") == "1" {
+			return 422, nil, &Error{422, "Cannot use `total' parameter: denied by configuration", nil}
+		}
+	}
+	q, e := route.Query()
+	if e != nil {
+		return e.Code, nil, e
+	}
+	list := &resource.ItemList{
+		Total: -1,
+		Items: []*resource.Item{},
+	}
+	var err error
+	if forceTotal {
+		list.Total, err = rsc.Count(ctx, q)
+		// If Storer doesn't implement Counter interface,
+		// fallback to listGet implementation
+		if err == resource.ErrNotImplemented {
+			return listGet(ctx, r, route)
+		}
+	}
+	if err != nil {
+		e = NewError(err)
+		return e.Code, nil, e
+	}
+	if win := q.Window; win != nil {
+		if win.Offset > 0 {
+			list.Offset = win.Offset
+		}
+		if win.Limit >= 0 {
+			list.Limit = win.Limit
+		}
+	}
+	return 200, nil, list
+}
--- a/rest/method_head_test.go
+++ b/rest/method_head_test.go
@@ -1,0 +1,56 @@
+package rest
+
+import (
+	"context"
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/rs/rest-layer-mem"
+	"github.com/rs/rest-layer/resource"
+	"github.com/rs/rest-layer/schema"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestHandlerHead(t *testing.T) {
+	s := mem.NewHandler()
+	s.Insert(context.TODO(), []*resource.Item{
+		{ID: "1"},
+		{ID: "2"},
+		{ID: "3"},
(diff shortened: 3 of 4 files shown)


resource/resource.go
```diff
@@ -342,6 +342,25 @@ func (r *Resource) find(ctx context.Context, q *query.Query, forceTotal bool) (l
 	return
 }
 
+// Count implements Storer interface.
```
rs: Count is not part of the Storer interface.

resource/resource.go
```diff
@@ -342,6 +342,25 @@ func (r *Resource) find(ctx context.Context, q *query.Query, forceTotal bool) (l
...
+		defer func(t time.Time) {
+			found := n
+			Logger(ctx, LogLevelDebug, fmt.Sprintf("%s.Count(...)", r.path), map[string]interface{}{
+				"duration": time.Since(t),
+				"found":    found,
+				"error":    err,
+			})
+		}(time.Now())
+	}
+	if err = r.hooks.onFind(ctx, q); err == nil {
```
rs: Should be an onCount handler or no handler.

resource/resource.go
```diff
@@ -342,6 +342,25 @@ func (r *Resource) find(ctx context.Context, q *query.Query, forceTotal bool) (l
...
+		n, err = r.storage.Count(ctx, &query.Query{Predicate: q.Predicate})
```
rs: How can this compile? I don’t see where you are adding `Count` to the storage type.

rest/method_head.go
```diff
@@ -0,0 +1,54 @@
...
+		if err == resource.ErrNotImplemented {
+			return listGet(ctx, r, route)
+		}
+	}
+	if err != nil {
+		e = NewError(err)
+		return e.Code, nil, e
+	}
+	if win := q.Window; win != nil {
+		if win.Offset > 0 {
```
rs: If offset is off bounds, shouldn’t we return a 404?

resource/resource.go
Dragomir-Ivanov: My bad, will fix.

resource/resource.go
Dragomir-Ivanov: `r.storage` is defined here:
```
type Resource struct {
...
	storage     storageHandler
...
}

type storageHandler interface {
	Storer
	MultiGetter
	Counter
	Get(ctx context.Context, id interface{}) (item *Item, err error)
}
```
But
```
type storageWrapper struct {
	Storer
}
```
Satisfies `storageHandler` and has `Count` function:
```
func (s storageWrapper) Count(ctx context.Context, q *query.Query) (total int, err error) {
	if s.Storer == nil {
		return -1, ErrNoStorage
	}
	if ctx.Err() != nil {
		return -1, ctx.Err()
	}
	if c, ok := s.Storer.(Counter); ok {
		return c.Count(ctx, q)
	}
	return -1, ErrNotImplemented
}
```

rest/method_head.go
Dragomir-Ivanov: Let me think a bit on Offsets.

resource/resource.go
Dragomir-Ivanov: As I understand it, HEAD must behave like GET, just not to get the result, so I reused the handlers. Having no handlers on HEAD will prevent implementing security policies, like preventing a request of getting total resource count, etc.

rest/method_head.go
Dragomir-Ivanov: There is explicit setting for Offset&Limit here. I think this is not right. Please lets discuss.
https://github.com/rs/rest-layer/blob/0df312935ab2631d1813920adf9bb04b328dd071/rest/routing.go#L198

resource/resource.go
smyrman: With the implementations you list though, should we not fallback to use Find with `Window{Limit=0}` when the Count interface is _not_ implemented, instead of returning an error?

rest/method_head.go
smyrman: @Dragomir-Ivanov, that is the _previous_ implementaton of HEAD, so I suppose you could change it _if_ you plan to include the offset/count as response headers. However, you _should_ probably call Find on the storage backend with these Window settings as a fallback when the `Counter` interface is not implemented.

smyrman: @rs, a bit of-topic for the review, but in the code that @Dragomir-Ivanov links to, I see you only handle `GET` and `HEAD`. I suppose that in order to fix #127, `page`, `limit` and `offset` (although not `fields`) should be handled in the case of `DELETE`.

resource/resource.go
smyrman: I would tend to agree with @Dragomir-Ivanov. There are other cases than _authentication_ hooks for the Find logic, such as manipulating the fields, but I suppose those would just work anyway if the response is empty.

Dragomir-Ivanov: I am comfortable with adding separate hooks for HEAD, as well. I think that gives the maximum flexibility. Users then can use shared code in `onFind/OnFound` and `onHead`. I guess no past-tense hook is needed, since there is no items to return.

resource/resource.go
Dragomir-Ivanov: @smyrman This could work, lets see what @rs will say.

rest/method_head.go
rs: @smyrman that's a good point.

resource/resource.go
Dragomir-Ivanov: @rs  any comment on this?

resource/resource.go
Dragomir-Ivanov: @rs Any comment on the hooks issue?

resource/resource.go
rs: Sounds good.

rest/method_head.go
Dragomir-Ivanov: @smyrman I think out-of-bounds offset doesn't work for current master:
```
GET http://192.168.0.103:3001/api/clients?skip=2&total=1&filter={"isBlacklisted":false}&limit=10

HTTP/1.1 200 OK
Content-Length: 660
Content-Type: application/json
Date: Wed, 18 Oct 2017 13:52:23 GMT
Request-Id: b7jlp5p40fm078fk87jg
Vary: Origin
X-Offset: 2
X-Total: 4

[
  {...},
  {...}
]
```

However this:
```
GET http://192.168.0.103:3001/api/clients?skip=100&total=1&filter={"isBlacklisted":false}&limit=10

HTTP/1.1 200 OK
Content-Length: 2
Content-Type: application/json
Date: Wed, 18 Oct 2017 13:54:20 GMT
Request-Id: b7jlq3140fm078fk87k0
Vary: Origin
X-Offset: 100
X-Total: 100

[]
```
It is buggy ( notice the value for `X-Offset`/`X-Total` ) and still returns `200 OK`.

Dragomir-Ivanov: Okay, I think that this specialization is not needed. In the recent brainstorm about the ETags, all items will need to be extracted ( or at least their `_etag` fields ) and be combined/summed in order to support `ETag` for list resources. Sorry about the noise. 
Let me understand what out-of-bound `offset` is, so I can fix at least this.

smyrman: I think, based on a very fast Google, the best thing for out of bounds, is to continue returning `200 OK`.

Regarding the `X-Total`, is it fixed in rs/rest-layer-mongo#19 in terms of the mongo storage backend at least?

Dragomir-Ivanov: Okay, so leaving it as it is, and ignoring Olivier's proposal for returning 404. 
@rs Are you okay with this?

smyrman: He is (see slack)

Dragomir-Ivanov: Closing this PR, since `200 OK` is current behavior, and having optimized HEAD implementation is not possible we will need to generate `ETag` based on DB contents.

Issue #148, opened 2017-10-03
Not filtering fields with multiple sub-resources


Dragomir-Ivanov:
Trying to get multiple sub-resources and selecting its fields on each of them, doesn't work as expected.

My concrete example:
```
GET http://192.168.0.103:3001/api/clients/59a962a5eaba9ebd0180ac67?fields=id,name,address,telephones,notes,visits,services{id,createdDT,status,name,zipCode,insuraceCompany,workOrder,ownOrder},appliances
```
, doesn't filter the fields of the `services`. If I remove the `appliances` sub-resource, it filters them.

Dragomir-Ivanov:
PR supplied #154

Dragomir-Ivanov:
Resolved by #154

PR #154, opened 2017-10-11
Resolver datarace

Dragomir-Ivanov:
This fixes nasty data-race within execution of `resourceBatchSolver`. #148 
When getting multiple sub-resources:
```
GET http://192.168.0.103:3001/api/clients/59a1a17aeaba9ebd0180abb0?fields=address,services{id,name,ownOrder,insuraceCompany},visits{name,appointmentDT,duration}
```

the last projection `{name,appointmentDT,duration}` is applied to all sub-resources (`services` and `visits`).
This was traced back to shared variable `pf` which was used in a loop, and reused by Golang `range` operator.

Didn't found in which test file to put this test case exactly. I can supply demo-app if requested.

### DIFF
--- a/schema/query/projection_evaluator.go
+++ b/schema/query/projection_evaluator.go
@@ -50,7 +50,8 @@ func evalProjection(ctx context.Context, p Projection, payload map[string]interf
 			p = append(p, ProjectionField{Name: fn})
 		}
 	}
-	for _, pf := range p {
+	for i := range p {
+		pf := p[i]
 		name := pf.Name
 		// Handle aliasing
 		if pf.Alias != "" {


schema/query/projection_evaluator.go
```diff
@@ -115,14 +115,14 @@ func evalProjection(ctx context.Context, p Projection, payload map[string]interf
...
 				}
-				rbr.request(ref.Path, q, func(payloads []map[string]interface{}, validator schema.Validator) (err error) {
+				rbr.request(ref.Path, q, &p[pidx], func(payloads []map[string]interface{}, validator schema.Validator, pf *ProjectionField) (err error) {
 					for i := range payloads {
 						if payloads[i], err = evalProjection(ctx, pf.Children, payloads[i], validator, rbr); err != nil {
 							return fmt.Errorf("%s: error applying projection on sub-resource item #%d: %v", pf.Name, i, err)
 						}
 					}
 					var v interface{}
-					if v, err = resolveFieldHandler(ctx, pf, def, payloads); err != nil {
```
rs: If I understand the data race correctly, you don't need to do all that to workaround it. You just need to copy the Projection at the beginning of the for loop:

Instead of:

```go
for _, pf := range p {
```

do:

```go
for i := range p {
    pf = p[i]
```

or less readable:

```go
for _, pf := range p {
    pf := pf
```

On a side note, the complexity of this function is thru the roof. We might want to break it down at some point.

Dragomir-Ivanov:
Done, please review.

Dragomir-Ivanov:
Merge maybe?

Issue #147, opened 2017-10-03
Returning whole resource, along with its sub-resouce


Dragomir-Ivanov:
Right now, having a query:
`api/shema/[id]` works and returns all fields without the referrers
`api/shema/[id]?fields=id,name,sub` works and returns `id` and `name` of the `shema` resource and its `sub` sub-resource.

The problem is that we can't have both. Would like to return all the fields of the schema without specifing them one by one. Something like:
`api/shema/[id]?fields=*,sub`

smyrman:
>  If I remove the appliances sub-resource, it filters them

This sounds like a bug. Do you mind raise this as a separate issue?

Dragomir-Ivanov:
@smyrman  In a second. Can you look at issue #123 and test this at your side. It might be real after all.

smyrman:
#148 raised to deal with the buggy part.

This issue will deal with how _expand_ one or more fields while still returning all fields without having to _list_ all fields.

Dragomir-Ivanov:
PR #155

Dragomir-Ivanov:
Guys I want to revive discussion on this topic. I did a PR #155 about it, however we didn't agreed on solution, so it got stale. I don't want anything fancy, just want to have a way to get all fields of resource/sub-resource. My motivation is as follows:
1. Development speed. It is more convenient to say `?fields=*,reference{*}` than manually adding each new field you add to resources during development sprints.
2. Since `?fields=reference{}` doesn't work, and `rest-layer` returns error `ParseProjection unexpected error: looking for field name at char 10`, client of the REST API doesn't have a way to know the shape of `reference` resource. This can be a real bumper, especially of lacking docs and if `reference` resource is not accessible from `index` but only used as sub-resource. 

I have a private patch that seems to work, and it allows queries like these:
```
?fields=*,reference{*}
?fields=*,n:name        // This will return name and n, as alias to name
?fields=reference{*}(filter:{name:"foo"}}
?fields=reference{*,user{*}}
```
I want to make this right, so no rush. In the meantime I will use mine patch.
Your input appreciated.

smyrman:
Just to clarify, I assume `reference` in your example is meant to be a field-name. With that assumption, I think your way of handling it makes sense to me.

I want to raise though, that I think the most useful use-case or a `*` expansion, would be to allow expanding references in a map where all the fields can't necessarily be known in advance. E.g. consider a field:

```go
"refmap": `{
    Validator: &schema.Dict{
[...]

Issue #138, opened 2017-09-18
Allow embedding to rely on an interface.


smyrman:
Today embedding is done by checking if a `schema.FieldValidator` (or fallback validator stored within a resource's index) type asserts to `&schema.Resource` or `&schema.Connection`. This makes it hard to write custom reference Validators and have them work just as well as the default ones. This could sometimes be useful when you have very application specific validation requirements for linked resources.

One solution to this, cold be to let `schema.Resource` and `schema.Connection` implement an interface each that let's you lookup the linked resource. E.g.

```go
type ReferenceType interface {
        LinkedResource() query.Resource
}
type ConnectionType interface {
        LinkedResource() (field string, query.Resource)
}
```

An alternate (and faster) implementation would be to return the `Path` instead of the `query.Resource`, but that might be less flexible, e.g. if we wanted to support exotic use-cases such as "external resources".

smyrman:
As an example of the latter use-case, and "external resource" could be a resource (e.g. "users") from a different REST-API that we somehow have managed to wrap into a query.Resource compatible type.

rs:
Sounds like a good idea.

smyrman:
Should probably deprecate `schema.ReferenceChecker`, and use `query.Resource` as the inteface for item extraction. https://github.com/rs/rest-layer/blob/48c11e50091e8c988780853a3aaa8af25cf98d42/schema/query/projection_evaluator.go#L12

Otherwise see in light of #77, #192, #194.

PR #155, opened 2017-10-11
Add a way to say "all fields" in the projection

Dragomir-Ivanov:
This enables us to get implicitly all fields of a resource, when sub-resources are requested #147 
```
GET http://192.168.0.103:3001/api/clients/59a1a17aeaba9ebd0180abb0?fields=*,services{ownOrder},visits{name,appointmentDT,duration}
```

This makes the following two requests equivalent:
```
GET http://192.168.0.103:3001/api/clients/59a1a17aeaba9ebd0180abb0?fields=*,services{*},visits{name,appointmentDT,duration}

GET http://192.168.0.103:3001/api/clients/59a1a17aeaba9ebd0180abb0?fields=*,services,visits{name,appointmentDT,duration}
```

### DIFF
--- a/schema/query/projection_evaluator.go
+++ b/schema/query/projection_evaluator.go
@@ -39,6 +39,27 @@ func (p Projection) Eval(ctx context.Context, payload map[string]interface{}, rs
 	return payload, err
 }
 
+func checkStar(p Projection) (bool, error) {
+	hasStar := false
+	s := ""
+	for i := range p {
+		if i > 0 {
+			s += ","
+		}
+		s += p[i].Name
+
+		if len(p[i].Children) == 0 {
+			if p[i].Name == "*" {
+				hasStar = true
+			}
+			if hasStar && i != 0 {
+				return false, fmt.Errorf("%s: invalid value: * must be the only field", s)
+			}
+		}
+	}
+	return hasStar, nil
+}
+
 func evalProjection(ctx context.Context, p Projection, payload map[string]interface{}, validator schema.Validator, rbr *referenceBatchResolver) (map[string]interface{}, error) {
 	res := map[string]interface{}{}
 	resMu := sync.Mutex{} // XXX use sync.Map once Go 1.9 is out
@@ -50,6 +71,18 @@ func evalProjection(ctx context.Context, p Projection, payload map[string]interf
 			p = append(p, ProjectionField{Name: fn})
 		}
 	}
+
+	isStar, err := checkStar(p)
+	if err != nil {
+		return nil, err
+	}
+	if isStar {
+		// When projection is star, it's like saying "all fields".
+		for fn := range payload {
+			p = append(p, ProjectionField{Name: fn})
+		}
+	}
+
 	for i := range p {
 		pf := p[i]
 		name := pf.Name
--- a/schema/query/projection_parser.go
+++ b/schema/query/projection_parser.go
@@ -178,7 +178,12 @@ func (p *projectionParser) scanFieldName() string {
 	field := []byte{}
 	for p.more() {
 		c := p.peek()
-		if (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_' || c == '-' {
+		if (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_' || c == '-' || c == '*' {
+			// Allow * only by itself.
+			if (c == '*' && len(field) != 0) || (c != '*' && len(field) > 1 && field[0] == '*') {
+				return ""
+			}
+
 			field = append(field, c)
 			p.pos++
 			continue
--- a/schema/query/projection_validator_test.go
+++ b/schema/query/projection_validator_test.go
@@ -34,32 +34,38 @@ func TestProjectionValidate(t *testing.T) {
 	}
 	cases := []struct {
 		projection string
-		err        error
+		parseErr   error
+		validErr   error
 	}{
-		{`parent{child},simple`, nil},
-		{`with_params(foo:1)`, nil},
-		{`with_params(bar:true)`, nil},
-		{`with_params(bar:false)`, nil},
-		{`with_params(foobar:"foobar")`, nil},
-		{`foo`, errors.New("foo: unknown field")},
-		{`simple{child}`, errors.New("simple: field as no children")},
-		{`parent{foo}`, errors.New("parent.foo: unknown field")},
-		{`simple(foo:1)`, errors.New("simple: params not allowed")},
-		{`with_params(baz:1)`, errors.New("with_params: unsupported param name: baz")},
-		{`with_params(foo:"a string")`, errors.New("with_params: invalid param `foo' value: not an integer")},
-		{`with_params(foo:3.14)`, errors.New("with_params: invalid param `foo' value: not an integer")},
-		{`with_params(bar:1)`, errors.New("with_params: invalid param `bar' value: not a Boolean")},
-		{`with_params(foobar:true)`, errors.New("with_params: invalid param `foobar' value: not a string")},
+		{`parent{child},simple`, nil, nil},
+		{`with_params(foo:1)`, nil, nil},
+		{`with_params(bar:true)`, nil, nil},
+		{`with_params(bar:false)`, nil, nil},
+		{`with_params(foobar:"foobar")`, nil, nil},
+		{`foo`, nil, errors.New("foo: unknown field")},
+		{`simple{child}`, nil, errors.New("simple: field as no children")},
+		{`parent{foo}`, nil, errors.New("parent.foo: unknown field")},
+		{`simple(foo:1)`, nil, errors.New("simple: params not allowed")},
+		{`with_params(baz:1)`, nil, errors.New("with_params: unsupported param name: baz")},
+		{`with_params(foo:"a string")`, nil, errors.New("with_params: invalid param `foo' value: not an integer")},
+		{`with_params(foo:3.14)`, nil, errors.New("with_params: invalid param `foo' value: not an integer")},
+		{`with_params(bar:1)`, nil, errors.New("with_params: invalid param `bar' value: not a Boolean")},
+		{`with_params(foobar:true)`, nil, errors.New("with_params: invalid param `foobar' value: not a string")},
+		{`*`, nil, nil},
+		{`parent,*`, nil, nil},
+		{`*,parent`, nil, nil},
+		{`*parent`, errors.New("looking for field name at char 2"), nil},
+		{`parent*`, errors.New("looking for field name at char 6"), nil},
 	}
 	for i := range cases {
 		tc := cases[i]
 		t.Run(tc.projection, func(t *testing.T) {
 			pr, err := ParseProjection(tc.projection)
-			if err != nil {
-				t.Errorf("ParseProjection unexpected error: %v", err)
+			if !reflect.DeepEqual(err, tc.parseErr) {
+				t.Errorf("ParseProjection error = %v, wanted: %v", err, tc.parseErr)
 			}
-			if err = pr.Validate(s); !reflect.DeepEqual(err, tc.err) {
-				t.Errorf("Projection.Validate error = %v, wanted: %v", err, tc.err)
+			if err = pr.Validate(s); !reflect.DeepEqual(err, tc.validErr) {
+				t.Errorf("Projection.Validate error = %v, wanted: %v", err, tc.validErr)
 			}
 		})
 	}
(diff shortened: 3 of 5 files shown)


schema/query/projection_evaluator.go
```diff
@@ -42,7 +42,7 @@ func (p Projection) Eval(ctx context.Context, payload map[string]interface{}, rs
 func evalProjection(ctx context.Context, p Projection, payload map[string]interface{}, validator schema.Validator, rbr *referenceBatchResolver) (map[string]interface{}, error) {
 	res := map[string]interface{}{}
 	resMu := sync.Mutex{} // XXX use sync.Map once Go 1.9 is out
-	if len(p) == 0 {
+	if len(p) == 0 || p[0].Name == "*" {
```
rs: What if * is supplied with other fields? Shouldn't we reject it?

schema/query/projection_parser.go
```diff
@@ -178,7 +178,7 @@ func (p *projectionParser) scanFieldName() string {
 	field := []byte{}
 	for p.more() {
 		c := p.peek()
-		if (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_' || c == '-' {
+		if (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_' || c == '-' || c == '*' {
```
rs: This will allow `*` in field names, which is not what we want right? The `*` should only be allowed by itself.

schema/query/projection_validator_test.go
```diff
@@ -50,6 +50,7 @@ func TestProjectionValidate(t *testing.T) {
 		{`with_params(foo:3.14)`, errors.New("with_params: invalid param `foo' value: not an integer")},
 		{`with_params(bar:1)`, errors.New("with_params: invalid param `bar' value: not a Boolean")},
 		{`with_params(foobar:true)`, errors.New("with_params: invalid param `foobar' value: not a string")},
+    {`*`, nil},
```
rs: Please add cases for:
* Fields with `*` in its name.
* `*` followed by valid fields.
* Valid fields followed by `*`.

Also, beware indentation. Are you using `gofmt`?

schema/query/projection_evaluator.go
Dragomir-Ivanov: I am not sure. Having `address,*` will have all fields `address` included. I can add a condition, where `*` to be the only predicate, else ignore it, but It doesn't feel right.

schema/query/projection_validator_test.go
Dragomir-Ivanov: I will add the cases. Indentation problem was due to resolving merge conflict with GitHub Web GUI.

schema/query/projection_evaluator.go
Dragomir-Ivanov: @smyrman can you comment on this as well.

schema/query/projection_parser.go
Dragomir-Ivanov: Yes, will fix this.

schema/query/projection_evaluator.go
smyrman: I am not completely sure I follow your comment @rs.

I thought the whole benefit of the syntax was that you could do `fields="*,sub1{},sub2{},hiddenField"`, to expand `sub1` and `sub2` and maybe explicitly include `hiddenField`, while still including all fields that where not explicitly mentioned.

Am I missing something?

smyrman: Or do you mean support `address,*` as well as `*,address`?

rs: I thought about `sub{*}` but not `*,sub{}`. The problem I think we have here is that `*` will already contain `sub` as an id. So we will end up with a duplicated field.

Did I miss something?

Dragomir-Ivanov: @smyrman Having sub-resource like `sub{}` is not possible at the moment. My patch will allow `sub{*}` which is equivalent of just saying `sub`

Dragomir-Ivanov: @smyrman 
`address,*` and `*,address` will result the same result as just `*`. I guess @rs was implying that we should rise an error on `*,address`/`address,*` cases, where I think they are completely valid.

rs: @Dragomir-Ivanov I meant `*,sub{<something>}`. My point is to say that you patch might not allow `*` with any other non aliased field as it would result in duplicate field definition. I don't think we check that anyway, but my point is that the result would be unpredictable.

To be more specific, if we have a schema with the field `a` as string and `b` as sub field. Asking for `*,b{*}` would be like asking for `a,b,b{*}`. Will the result contain `b` as an id or `b` as a sub struct is unclear and depend on the implementation.

Dragomir-Ivanov: Okay, think I am missing something. Don't understand what you are explaining, and it is because I don't know `rest-layer` that well. Are you talking about nested objects within a resource ( `schema.Object` )

rs: Yes

Dragomir-Ivanov: Still can't quite understand the purpose of `schema.Object` and `schema.Connection` but I tried the following:
```
			"a": {
				Validator: &schema.String{},
			},
			"b": {
				Schema: &schema.Schema{
					Fields: schema.Fields{
						"a": {
							Validator: &schema.String{},
						},
						"b": {
							Validator: &schema.String{},
						},
					},
				},
			},
```
This is the data in the resource:
```
{
  "a": "Data 1",
  "b": {
    "a": "Data 2",
    "b": "Data 3"
  },
  "created": "2017-10-13T12:38:34.399542387+03:00",
  "id": "b7g8j6h40fm2fhph21ag",
  "name": "EMA",
  "updated": "2017-10-13T12:38:34.399543697+03:00"
}
```
All these requests:
```
GET http://192.168.0.103:3001/api/users/b7g8jpp40fm2fhph21eg?fields=*
GET http://192.168.0.103:3001/api/users/b7g8jpp40fm2fhph21eg?fields=*,b
GET http://192.168.0.103:3001/api/users/b7g8jpp40fm2fhph21eg?fields=*,b{*}
GET http://192.168.0.103:3001/api/users/b7g8jpp40fm2fhph21eg?fields=*,b{*},b{*}
GET http://192.168.0.103:3001/api/users/b7g8jpp40fm2fhph21eg?fields=*,b{*},b{*},*
and even
GET http://192.168.0.103:3001/api/users/b7g8jpp40fm2fhph21eg?fields=*,b{a}     This is because * has precedence  
```
Result the same response:
```
{
  "a": "Data 1",
  "b": {
    "a": "Data 2",
    "b": "Data 3"
  },
  "created": "2017-10-13T12:39:51.670839057+03:00",
  "id": "b7g8jpp40fm2fhph21eg",
  "name": "EMA",
  "updated": "2017-10-13T12:39:51.67084017+03:00"
}
```
Trying something different:
```
GET http://192.168.0.103:3001/api/users/b7g8jpp40fm2fhph21eg?fields=a,b{a}
```
Gets expected result:
```
{
  "a": "Data 1",
  "b": {
    "a": "Data 2"
  }
}
```

smyrman: When I talk abut sub-resources and hidden fields, I was thinking more like this:
```go
resourceA := index.Bind("parent", schema.Schema{
    Fields: {
        "showMe": {},
        "hiddenField": {Hidden: true},
}, ...)
rscParent := index.Bind("Parent", schema.Schema{
    Fields: {
        "showMe": {},
        "hiddenField": {Hidden: true},
    },
}, storeA, conf)
// afterc calling rscA.Bind, "sub" will now appear as a `schema.Connection` in `rscA` on a special _field override list_.
rscA.Bind("sub", "parent", schema.Schema{
    Fields: {
        "parent": {Validator: &schema.Reference{Path:"parent"}},
        "a": {},
        "hiddenField": {Hidden: true},
    },
})
```

And then ask for either `/parent?fields=*,sub{hiddenField}` or /parent?fields=sub{hiddenField},*`

So I guess `@rs` point is, if this is expanded to either `fields=showMe,sub,sub{a}` or  `fields=sub{a},showMe,sub`, which fields will `sub` really show? Clearly our intention of the user would be to display only `hiddenField`, but what will happen?

The bet approach for this discussion is probably to first add some tests for this, and then see what changes is needed for the implementaiton in order for the user assumption to hold.

smyrman: Btw, The test you showed us was really nice @Dragomir-Ivanov. But I am not sure I would _want_ `*` to have precedence...

rs: This asks another question too, should `fields=foo,foo` or `fields=f:foo,f:bar` raise an error or just result in a "last win"?

Dragomir-Ivanov: @rs Good question, I guess we should just pick one, and keep it until it proves wrong. I vote for `last win`.

rs: That’s the current implementation. We should add tests and document it then.

schema/query/projection_evaluator.go
```diff
@@ -39,6 +39,27 @@ func (p Projection) Eval(ctx context.Context, payload map[string]interface{}, rs
 	return payload, err
 }
 
+func checkStar(p Projection) (bool, error) {
+	hasStar := false
+	s := ""
+	for i := range p {
```
smyrman: This for loop is a bit awkward...

Could not this whole test be just:

```go
func checkStar(p Projection) (bool, error) {
    if len(p) == 1 && p[0].Name == "*" && len p[0].Children == 0 {
        return true, nil
    }
    if len(p) <= 1 {
        return false, nil
    }
    var s, comma string
    for i := range p {
        if p[i].Name == "*"  && len p[i].Children == 0: {
            return false, fmt.Errorf("%s: invalid value: * must be the only field", s)
        }
    }
}
```

This gives less indentation, and a slightly more clear LOS ([Line of Sight](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88)).

smyrman: An even smaller version with different errors...
```go
func checkStar(p Projection) bool {
    if len(p) == 1 && p[0].Name == "*" && len p[0].Children == 0 {
        return true
   }
   return false
}
```

smyrman: Otherwise I am now OK with `*` only being allowed as the only parameter, as long as this now expands all references; at least initially. May place nice (end-user-wise) with the suggesrion of using `*` to expand references in schema.Dict and schema.Array

#138

Dragomir-Ivanov: Thanks for the suggestions @smyrman. However expanding referenced fields, is what I wanted to avoid in the first place. My indention was to provide extension to this requests:
```
GET http://192.168.0.103:3001/api/clients/59a85126eaba9ebd0180ac55
```
Which will return only non referenced fields, so making this request:
```
GET http://192.168.0.103:3001/api/clients/59a85126eaba9ebd0180ac55?fields=*,visits
```
Is to actually add the `visits` referenced resource, without mentioning all the non referenced fields.

Initially I though this as a low hanging fruit, but it gets more complicated. It is not big deal to enumerate all the fields explicitly, so I think I will just abandon this PR.

smyrman: >  It is not big deal to enumerate all the fields explicitly, so I think I will just abandon this PR.

Alright 👍 Might be wise for now. This is also the way we use it.

Just for my clarification though.

```
GET http://192.168.0.103:3001/api/clients/59a85126eaba9ebd0180ac55?fields=*,visits

```
> Is to actually add the visits referenced resource, without mentioning all the non referenced fields.

This is also what I though was the initial intent. However, my (perhaps wrongful) understanding of the final PR was that it would expand "visits", all other references _and_ maybe (not clear to me) all fields that are hidden by default (Field.Hidden == true) if you did this:

```
GET http://192.168.0.103:3001/api/clients/59a85126eaba9ebd0180ac55?fields=*
```

If we rather want the initially described behaviour, I suppose the following might work:

```go
// expandStar replace a single `*` in projection with all fields at the position of `*`.
// duplicated fields will be handled according to the last-win principal.
expandStart(p *Projection, schema.Fields) {
    ...
}
```
And then no other changes.

Dragomir-Ivanov:
```
GET http://192.168.0.103:3001/api/clients/59a85126eaba9ebd0180ac55?fields=*
```
The request above would have been equivalent to this:
```
GET http://192.168.0.103:3001/api/clients/59a85126eaba9ebd0180ac55
```
No referenced fields, no hidden fields.
Actually there is no way of getting the hidden fields with the current code. Mentioning them explicitly in the `fields=` parameter results in the following error:
```
GET http://192.168.0.103:3001/api/clients/59a85126eaba9ebd0180ac55?fields=nameTokens
"Invalid `fields` parameter: nameTokens: hidden field"
```

```
Client = schema.Schema{
		Fields: schema.Fields{
			"id": mongo.ObjectIDField,
...
			"nameTokens": {
				Hidden: true,
			},
```

Dragomir-Ivanov: Closing the PR now.

smyrman: Thanks for your effort, even if it was not merged this time.

PR #159, opened 2017-10-19
Use "Skip" even if no "Limit" is specified

Dragomir-Ivanov:
Without this patch, query like this:
```
GET http://192.168.0.103:3001/api/clients?skip=1&filter={"isBlacklisted":false}
```
`skip` parameter will be unused, because there is no `limit` parameter specified. I view this as non-consistent behavior, so this patch fixes that.


Dragomir-Ivanov:
@smyrman Please review, and maybe merge.

PR #161, opened 2017-10-23
routing: HEAD must behave just like GET, except returning the body

Dragomir-Ivanov:
Fix HTTP HEAD request to respect `Window` parameters, and to make it function just like GET.

### DIFF
--- a/rest/routing.go
+++ b/rest/routing.go
@@ -195,10 +195,7 @@ func (r *RouteMatch) Query() (*query.Query, *Error) {
 	}
 	// Parse query string params.
 	switch r.Method {
-	case "HEAD":
-		// request empty response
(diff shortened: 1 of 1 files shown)


rest/routing.go
```diff
@@ -196,8 +196,7 @@ func (r *RouteMatch) Query() (*query.Query, *Error) {
 	// Parse query string params.
 	switch r.Method {
 	case "HEAD":
-		// request empty response
-		q.Window = &query.Window{Offset: 0, Limit: 0}
+		fallthrough
```
rs: Use `case "HEAD", "GET":` instead.

Dragomir-Ivanov:
Yes, thanks @smyrman.

Issue #157, opened 2017-10-13
ETag validity on composite resource


Dragomir-Ivanov:
Lets suppose we have this request:
```
GET http://192.168.0.103:3001/api/clients/59a85126eaba9ebd0180ac55?fields=*,services,visits
If-None-Match: "11b22a28b07ddf482f8269050e9691d9"
```
If I execute this in a row, first time I get 200 then 304, which is as it is supposed to be.
But then I change some field values of `services` and `visits` behind the scenes.

Then retry the query above, I still get 304, although the result of the query IS CHANGED.

ETag comparison has its place on updating concrete resources, but I think we should leave generation of ETags for compound resources, and for resouseLists to be generated in a middleware function executed after `rest-layer`.

rs:
Do you mean `*,services{*},visits{*}`? Because without `{}` you should just get the object ids.

This is a very valid point. What about computing a composite ETag?

cc @smyrman

Dragomir-Ivanov:
@rs `services` and `visits` are bound to `clients` by `schema.Reference` and not `schema.Connection` so `*,services,visits` results in an arrays of `services:[]` and `visits:[]` attached to the returned clients item.

Composite ETag can't be cached in the DB, so we need to recompute it on the fly, compare with the supplied ETag and return 304 if they match. 
Is all this work related to `rest-layer` at all, aren't we putting too many responsibilities on it? Other frameworks will do this in a middle-ware function executed after the frameworks.

I guess if we have some context ( but not Go context.Context) exported in the request, and we put some important values there to let the following middle-ware functions make decisions based on them.
I am against `context.Context` because it is hierarchical, and adding values in a child `ctx` doesn't propagate to parent `ctx`.

Dragomir-Ivanov:
Okay, I think now I am understanding why middle-ware path is not considered here. In Go `http.ResponseWriter` is single shot, and once you `Write()` there is no turning back. That is why `Alice` middle-ware can be used only for modifying the `http.Request` object, but not the response.
Please confirm that my understanding is correct.

rs:
Yes. That’s why you can set a custom ResponseSender on `rest.Handler`.

Computing a checksum of all the Etags shouldn’t be too expensive.

Dragomir-Ivanov:
@rs Thanks, then custom` ResponseSender` is the way to go.

> Computing a checksum of all the Etags shouldn’t be too expensive.
[...]

PR #162, opened 2017-10-23
Return weak ETags and add ETag HTTP header to itemList response

Dragomir-Ivanov:
Pickup the long-hanging fruit, and add `ETag` to itemList response.

### DIFF
--- a/rest/response.go
+++ b/rest/response.go
@@ -91,9 +91,17 @@ func (f DefaultResponseFormatter) FormatList(ctx context.Context, headers http.H
 	if l.Offset > 0 {
 		headers.Set("X-Offset", strconv.Itoa(l.Offset))
 	}
+
+	hash := md5.New()
+	for _, item := range l.Items {
+		if item.ETag != "" {
+			hash.Write([]byte(item.ETag))
+		}
+	}
+	headers.Set("ETag", `W/"`+fmt.Sprintf("%x", hash.Sum(nil))+`"`)
+
 	if !skipBody {
 		payload := make([]map[string]interface{}, len(l.Items))
-		hash := md5.New()
 		for i, item := range l.Items {
 			// Clone item payload to add the etag to the items in the list.
 			d := map[string]interface{}{}
@@ -102,11 +110,9 @@ func (f DefaultResponseFormatter) FormatList(ctx context.Context, headers http.H
 			}
 			if item.ETag != "" {
 				d["_etag"] = item.ETag
-				hash.Write([]byte(item.ETag))
 			}
 			payload[i] = d
 		}
-		headers.Set("ETag", `W/"`+fmt.Sprintf("%x", hash.Sum(nil))+`"`)
 		return ctx, payload
 	}
 	return ctx, nil
--- a/rest/response.go
+++ b/rest/response.go
@@ -2,6 +2,7 @@ package rest
 
 import (
 	"context"
+	md5 "crypto/md5"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -71,7 +72,7 @@ func (s DefaultResponseSender) Send(ctx context.Context, w http.ResponseWriter, 
 // FormatItem implements ResponseFormatter.
 func (f DefaultResponseFormatter) FormatItem(ctx context.Context, headers http.Header, i *resource.Item, skipBody bool) (context.Context, interface{}) {
 	if i.ETag != "" {
-		headers.Set("Etag", `"`+i.ETag+`"`)
+		headers.Set("Etag", `W/"`+i.ETag+`"`)
 	}
 	if !i.Updated.IsZero() {
 		headers.Set("Last-Modified", i.Updated.In(time.UTC).Format("Mon, 02 Jan 2006 15:04:05 GMT"))
@@ -92,6 +93,7 @@ func (f DefaultResponseFormatter) FormatList(ctx context.Context, headers http.H
 	}
 	if !skipBody {
 		payload := make([]map[string]interface{}, len(l.Items))
+		hash := md5.New()
 		for i, item := range l.Items {
 			// Clone item payload to add the etag to the items in the list.
 			d := map[string]interface{}{}
@@ -100,9 +102,11 @@ func (f DefaultResponseFormatter) FormatList(ctx context.Context, headers http.H
 			}
 			if item.ETag != "" {
 				d["_etag"] = item.ETag
+				hash.Write([]byte(item.ETag))
 			}
 			payload[i] = d
 		}
+		headers.Set("ETag", `W/"`+fmt.Sprintf("%x", hash.Sum(nil))+`"`)
 		return ctx, payload
 	}
 	return ctx, nil
--- a/rest/response_test.go
+++ b/rest/response_test.go
@@ -82,7 +82,7 @@ func TestDefaultResponseFormatterFormatItem(t *testing.T) {
 
 	h = http.Header{}
 	rctx, payload = rf.FormatItem(ctx, h, &resource.Item{ETag: "1234"}, false)
-	assert.Equal(t, http.Header{"Etag": []string{`"1234"`}}, h)
+	assert.Equal(t, http.Header{"Etag": []string{`W/"1234"`}}, h)
 	assert.Equal(t, rctx, ctx)
 	assert.Equal(t, map[string]interface{}(nil), payload)
 }
@@ -95,7 +95,7 @@ func TestDefaultResponseFormatterFormatList(t *testing.T) {
 		Total: -1,
 		Items: []*resource.Item{{Payload: map[string]interface{}{"foo": "bar"}}},
 	}, false)
-	assert.Equal(t, http.Header{}, h)
+	assert.Equal(t, http.Header{"Etag": []string{`W/"d41d8cd98f00b204e9800998ecf8427e"`}}, h)
 	assert.Equal(t, rctx, ctx)
 	assert.Equal(t, []map[string]interface{}{{"foo": "bar"}}, payload)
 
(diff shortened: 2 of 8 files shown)


rest/response.go
```diff
@@ -92,6 +93,7 @@ func (f DefaultResponseFormatter) FormatList(ctx context.Context, headers http.H
 	}
 	if !skipBody {
 		payload := make([]map[string]interface{}, len(l.Items))
+		etags := make([]byte, len(l.Items)*md5.Size)
```
rs: Would be better to create a new md5 hash instance and append etags to it instead (incremental hash) instead of storing all the etags in memory to then create a hash.

rest/response_test.go
```diff
@@ -95,7 +95,7 @@ func TestDefaultResponseFormatterFormatList(t *testing.T) {
 		Total: -1,
 		Items: []*resource.Item{{Payload: map[string]interface{}{"foo": "bar"}}},
 	}, false)
-	assert.Equal(t, http.Header{}, h)
+	assert.Equal(t, http.Header{"Etag": []string{"W/\"d41d8cd98f00b204e9800998ecf8427e\""}}, h)
```
smyrman: IF you use back ticks, you don't need to escape the double quotes:
```
[]string{`W/"d41d8cd98f00b204e9800998ecf8427e`}
```

rest/response.go
```diff
@@ -100,9 +102,11 @@ func (f DefaultResponseFormatter) FormatList(ctx context.Context, headers http.H
 			}
 			if item.ETag != "" {
 				d["_etag"] = item.ETag
+				hash.Write([]byte(item.ETag))
 			}
 			payload[i] = d
 		}
+		headers.Set("ETag", `W/"`+fmt.Sprintf("%x", hash.Sum(nil))+`"`)
```
smyrman: I would suggest doing a strong E-tag in the `ResponseSender` instead:

https://github.com/rs/rest-layer/blob/2c4c847a7e4d13e1245c93990665aaeb1a35cfe2/rest/response.go#L57

Then we can just do `hash.Write(j)` in one line - it's not affected by some of the resource embedding would give us, as discussed in #157.

If we are worried about the performance, I suggest we benchmark it with a few different response sizes. If it's just a few milliseconds, or less, I don't think we should worry.

rest/response.go
```diff
@@ -71,7 +72,7 @@ func (s DefaultResponseSender) Send(ctx context.Context, w http.ResponseWriter,
 // FormatItem implements ResponseFormatter.
 func (f DefaultResponseFormatter) FormatItem(ctx context.Context, headers http.Header, i *resource.Item, skipBody bool) (context.Context, interface{}) {
 	if i.ETag != "" {
-		headers.Set("Etag", `"`+i.ETag+`"`)
+		headers.Set("Etag", `W/"`+i.ETag+`"`)
```
smyrman: @rs -- for the latest idea of root-item E-tag + strong E-tag from #157 -- should we set it (as weak) hear, and then override it in the `DefaultResponseSender` to be:

```go
rootEtag := strings.StripPtefix(headers.Get("Etag"), "W/")
if rootEtag != "" {
    headers.Set("Etag", rootEtag + '-' + md5Etag)
} else {
    headers.Set("Etag", md5Etag)
}
```

rest/response.go
Dragomir-Ivanov: Lets see what @rs will say. For me, hash of hashes in this case is even stronger ETag, because included `_etag` field will be different even for items fields changed outside the projection.

smyrman: Actually... it appears the list-view `E-tag` is only include if `skipBody == false`. Will the `E-tag` here even be included on `HEAD` requests?

Maybe not so easy to do this Hash in the sender neither btw, as it doesn't get passed in the `skipBody` parameter.

rest/response.go
rs: Why not. We could event only do the MD5 if we detect cross resource projections.

rest/response.go
Dragomir-Ivanov: Yes, FormatList is the right place.

smyrman: > Lets see what @rs will say. For me, hash of hashes in this case is even stronger ETag, because included _etag field will be different even for items fields changed outside the projection.

Well, there are _no_ includede _etag fields atm, except for the list-view root-resources, and then it doesn't matter. Anyway, as pointed out, not trivial to do it in the sender -- unless we want to change the interfaces.

rest/response.go
smyrman: >  if we detect cross resource projections.

How do we detect that on this level though:-) It looks like supporting this would require some interface changes... Maybe leave it as is for now? And see if there is a good way to do it later?

rest/response.go
smyrman: For now, maybe just leave it as a weak E-tag of tags, but move it out of the `!if skipBody`:

```go
hash := md5.New()
for i, item := range l.Items {
        if item.Etag != "" {
                hash.Write([]byte(item.ETag))
        }
}

```

@rs, what do you think about that? We can look at handling embedded sub-resources and potentially strong E-tags later if we want to.

rest/response.go
rs: ok

Dragomir-Ivanov: Uploaded changes suggested. @smyrman Please review.

smyrman:
```go
--- FAIL: TestDefaultResponseFormatterFormatList (0.00s)
	Error Trace:	response_test.go:107
	Error:      	Not equal: 
	            	expected: http.Header{}
	            	actual: http.Header{"Etag":[]string{"W/\"d41d8cd98f00b204e9800998ecf8427e\""}}
	            	
	            	Diff:
	            	--- Expected
	            	+++ Actual
	            	@@ -1,2 +1,5 @@
	            	-(http.Header) {
	            	+(http.Header) (len=1) {
	            	+ (string) (len=4) "Etag": ([]string) (len=1) {
	            	+  (string) (len=36) "W/\"d41d8cd98f00b204e9800998ecf8427e\""
	            	+ }
	            	 }
	Error Trace:	response_test.go:117
	Error:      	Not equal: 
	            	expected: http.Header{"X-Total":[]string{"1"}, "X-Offset":[]string{"2"}, "Etag":[]string{"W/\"d41d8cd98f00b204e9800998ecf8427e"}}
	            	actual: http.Header{"X-Offset":[]string{"2"}, "Etag":[]string{"W/\"d41d8cd98f00b204e9800998ecf8427e\""}, "X-Total":[]string{"1"}}
	            	
	            	Diff:
	            	--- Expected
	            	+++ Actual
	            	@@ -2,3 +2,3 @@
	            	  (string) (len=4) "Etag": ([]string) (len=1) {
	            	-  (string) (len=35) "W/\"d41d8cd98f00b204e9800998ecf8427e"
	            	+  (string) (len=36) "W/\"d41d8cd98f00b204e9800998ecf8427e\""
	            	  },
```

Dragomir-Ivanov:
@smyrman  Fixed up.

Dragomir-Ivanov:
Yes, thanks @smyrman .

smyrman:
@rs, I can rebase & merge -- just updating against master first.

rest/util.go
```diff
@@ -126,11 +126,13 @@ func compareEtag(etag, baseEtag string) bool {
...
 		return false
 	}
-	if etag == baseEtag {
-		return true
-	}
-	if l := len(etag); l == len(baseEtag)+2 && l > 3 && etag[0] == '"' && etag[l-1] == '"' && etag[1:l-1] == baseEtag {
-		return true
+	if strings.HasPrefix(etag, "W/") {
+		if etag[2:] == baseEtag {
+			return true
```
rs: You are ignoring the quotes. This will never match.

smyrman:
Quotes come after `W/"<etag>"`  but do you mean the `baseEtag` lack the quotes?

Sorry for the premature merge

rest/util.go
smyrman: I see what you mean now! @Dragomir-Ivanov, do you want to do a quick PR where you handle the quotes?

Dragomir-Ivanov:
I think that quotes are optional for `rest-layer`. Thanks next condition will match the qoutes as well.

rest/util.go
```diff
@@ -126,11 +126,13 @@ func compareEtag(etag, baseEtag string) bool {
...
+		}
+		if l := len(etag); l == len(baseEtag)+4 && l > 4 && etag[2] == '"' && etag[l-1] == '"' && etag[3:l-1] == baseEtag {
+			return true
+		}
```
smyrman: > I think that quotes are optional for rest-layer. Thanks next condition will match the qoutes as well.

Right. Maybe we should simplify the code here a bit to make it more readable.

```go
if etag == "" {
	return false
}
cmpEtag := strings.StripPrefix(etag, "W/")
cmpEtag = strings.StripPrefix(etag, `"`)
cmpEtag = strings.StripSuffix(etag, `"`)
return cmpEtag == baseEtag
``

smyrman: PS! You don't need to do this if you don't feel like it. Maybe I can do it ;-) I kind of want to benchmark it as well, just for fun.

Dragomir-Ivanov: Yes, I guess we can simplify it a little more. I was just following already written code style.

I don't expect any significant overhead of computing the ETags, compared to actual network latency. You can write some benchmarks of you want though.

smyrman: > . I was just following already written code style.

It's OK, don't worry abuout it. It was nice work!

The benchmarks would only for my own case, and just for fun. It's easy to assume one thing is faster than another without being right, and this things migh affect the code one write in general.

I think we should do the cleanups either way, but I can take care of it.

PR #163, opened 2017-11-02
response: Fix _etag formatting in list items

Dragomir-Ivanov:


rs:
Why do we need to do that in payload? The weak etag format is only relevant when used as an HTTP header.

smyrman:
> Why do we need to do that in payload? The weak etag format is only relevant when used as an HTTP header.

Valid point; there is no need to worry about the _etag in the payload being weak or not. I suppose we can close this review then.

smyrman:
@Dragomir-Ivanov, any thoughts?

Dragomir-Ivanov:
Yes, makes sense. Since this is actual resource ETag, then it must be strong.

PR #169, opened 2017-11-16
Fix schema.Dict.GetField bug

smyrman:
fixes GetField(string) to work when KeysValidator is not set, and ads
tests for it. Using asserts library for consistent output compared
to existing tests.

### DIFF
--- a/schema/dict_test.go
+++ b/schema/dict_test.go
@@ -3,6 +3,7 @@
 package schema_test
 
 import (
+	"reflect"
 	"testing"
 
 	"github.com/rs/rest-layer/schema"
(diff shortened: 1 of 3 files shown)


schema/dict_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/rs/rest-layer/schema"
+	"github.com/stretchr/testify/assert"
```
rs: Can we stop using testify for new tests?

smyrman: Of course, but I don't see what you are trying to acheive by dropping it.

I have been writing my share off go tests withou a matcher library, but it gives you nothing for free.

Especially when it come to JSON, struct or byte array comparison, `testify/assert` and `testify/require` gives you a free diff of the two items, showing you what's wrong; that doesn't just make writing the test faster, it makes debugging issues faster.

What's the strateg for something like that without a matcher librarys? Our own internal matcher functions? And does that make it any better or more readabld to third-parties?

rs: I got beat by testify’s sloppy comparison when used on `nil` interfaces for instance. I also tend to prefer standard go test constructs, it’s less work for new contributors to read our tests.

What about using `google/go-cmp` for struct comparisons but move to pure go constructs?

smyrman: Cool. There is a few struct/JSON comparison tools. I haven treid `google/go-cmp` but I will have a look at it. For these tests I will transfer them to use `reflect.DeepEqual` as the diff is less relevant.

> I got beat by testify’s sloppy comparison when used on nil interfaces for instance

Nil interfaceses in Go is a bad one.

smyrman:
@rs, tests updated.

PR #172, opened 2017-11-23
Avoid rest package calling OnFind/OnFound hooks on PATCH

### LINKED ISSUE #171: Document behavior of resource hooks when called from the rest package
When sending a PATCH request to the `rest` package, OnFound resource hooks get's triggered. This can cause issues when:

- The OnFound hook adds fields or does changes too the response that should not be written.
- The OnFound hook is expensive.

https://github.com/rs/rest-layer/blob/326f8072a1edb7738b3e4bed8a573a0c2ffdfe29/rest/method_item_patch.go#L27

There might be valid use-cases for running the OnFound hook on PATCH, e.g. when the hooks manages an on-read data migration, however this is still best done explicitly through an OnUpdate hook that migrates `original`.

## Solution

There is a simple enough solution. Since we are using context.Context we can use it to signal the resource package to skip hooks. E.g.

```go
// In package resource somewhere.
DisableHooks(ctx context.Context) context.Context {
    // ...
}

// In the linked rest code:
 if l, err := rsrc.Find(resource.DisableHooks(ctx), q); err != nil { 
```
Update all places where hooks gets called to check for the `ctx.Value` set by `DisableHooks` .

smyrman:
Fixes #171, and updates other intermediate calls to the resource package done by the rest package, so that they do not trigger unintended hooks.

### DIFF
--- a/rest/method_item_patch.go
+++ b/rest/method_item_patch.go
@@ -24,7 +24,7 @@ func itemPatch(ctx context.Context, r *http.Request, route *RouteMatch) (status 
 	rsrc := route.Resource()
 	var original *resource.Item
 	q.Window = &query.Window{Limit: 1}
-	if l, err := rsrc.Find(ctx, q); err != nil {
+	if l, err := rsrc.Find(resource.WithDisableHooks(ctx), q); err != nil {
 		// If item can't be fetch, return an error.
 		e = NewError(err)
 		return e.Code, nil, e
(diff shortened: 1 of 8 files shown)


rest/method_item_patch.go
```diff
@@ -37,7 +37,7 @@ func itemPatch(ctx context.Context, r *http.Request, route *RouteMatch) (status
 	if err := checkIntegrityRequest(r, original); err != nil {
 		return err.Code, nil, err
 	}
-	changes, base := rsrc.Validator().Prepare(ctx, payload, &original.Payload, false)
+	changes, base := rsrc.Validator().Prepare(resource.WithDisableHooks(ctx), payload, &original.Payload, false)
```
smyrman: not really needed here... Added it more as a precaution than anything else...

smyrman:
I am happy with another API being considered, but it's a hard, at least for me, to come up woth a better compremice. Do you have any good sugestion that solves the problem? Add Patch to the resource maybe? Or make calling the hooks an explicit call?

I don't quite buy that it's ineficient without a benchmark though, and I don't see how it's fragile? It appears in line with the context.Context use-case of a request-scoped flag (although I am not a huge fan of the context.Context mixed-pot cancel + random variables idea).

rs:
Context is ok for carrying contextual data like auth user. It should not be used to carry hidden parameters for a function.

I think the solution, as describe in the other ticket, is to step back for a sec, and see if the use case you have shouldn’t be solved a different way.

Did you try to simply add your virtual field to the schema?

smyrman:
Closing PR, as this is not the right approach.

smyrman:
> Did you try to simply add your virtual field to the schema?

as an FYI, yes. More info on the ticket.

Issue #176, opened 2018-02-28
Filtering on fields of type array does not work


smyrman:
This is the same issue as for #139, but for arrays.

1. Filtering agains an array with exact match fails due to the use of `query.Value`, as an array in the filter is parsed to `[]query.Value` and not `[]interface{}`
2. Filtering agains array indices (`myarray.0: "a", myarray.1: "b"`) does not work.

Case 1 should be solved together for both arrays and dict.

Case 2 should be solvable in much the same way as it was solved for #139. I will start with that now.

smyrman:
Case 1 solved by #177, so only Case 2 remain

smyrman:
Case 2 solved by #179

PR #177, opened 2018-02-28
Make query.Value a type alias of interface{}

torie:
This fixes the case where filtering for dictionary keys or arrays fails due to a mismatch of types as described in https://github.com/rs/rest-layer/issues/139 and https://github.com/rs/rest-layer/issues/176.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-- 1.7
+- 1.9
 - tip
 matrix:
   allow_failures:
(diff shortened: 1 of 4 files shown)


smyrman:
@torie, was there also some translating methods that can be removed?

Travis tests must be updated to run a later version of Go as type aliases where introduced in Go 1.8. I suggest Go 1.10, and/or plausibly 1.9.

torie:
If this is merged, the `valuesToInterface()` conversion can also be removed from the storage backends where it's used, such as here https://github.com/rs/rest-layer-mongo/blob/4b10584d0513745d36ba07e93d3acc8f92323231/query.go#L90.

PR #179, opened 2018-02-28
Array support for index lookup (breaking change)

smyrman:
This commit adds support for looking up field definitions inside
schema.Array instances. To achieve this, Array.ValuesValidator has been
deprecated in favour of a new Array.Values property that holds a full
schema.Field.

As a result, it should be possible to filter on schema.Array index value
in backends that supports it.

TODO:
- [x] Implement changes 
- [x] Verify against MongoDB backend
- [x] Update README


smyrman:
@rs, ready for review.

Issue #153, opened 2017-10-11
Storer interface documentation out of date


smyrman:
From the README:

```go
type Storer interface {
	Find(ctx context.Context, lookup *resource.Lookup, offset, limit int) (*resource.ItemList, error)
	Insert(ctx context.Context, items []*resource.Item) error
	Update(ctx context.Context, item *resource.Item, original *resource.Item) error
	Delete(ctx context.Context, item *resource.Item) error
	Clear(ctx context.Context, lookup *resource.Lookup) (int, error)
}
```

PR #181, opened 2018-03-03
Resolves rs/rest-layer#153 linted README Markdown.

SAFAD:


rs:
Go code uses tab for indentation. Why do you want to use space instead?

smyrman:
Actually, the README is using a mix if tabs and spaces atm, also for (Go) code examples, probably due to being copy-pasted in from various sources.

I don't mind which is used for Markdown, spaces or tabs, but it's nice to make it consistent. Sicking with some linter's default option may be the road of shortest resistance. @SAFAD, which linter did you use?

smyrman:
> Actually, the README is using a mix if tabs and spaces atm, also for (Go) code examples

Actually, Go code seams to be mostly using tabs, while terminal example code is using spaces in master atm.

SAFAD:
@rs @smyrman Yes I know that Go indentation is tab based, however [markdownlint](https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md) kept whining about it.

We can use the following to make the linter ignore the Hard tabs in Go code:

> ```<!-- markdownlint-disable MD010 -->```
> ```go
> users := index.Bind("users", user, mem.NewHandler(), resource.Conf{
> 	AllowedModes: []resource.Mode{resource.Read, resource.List, resource.Create, resource.Delete},
> })
> ```
> ```<!-- markdownlint-enable MD010 -->```

Shall I push a new commit to match this style for each Go code block ?

smyrman:
I would vote for using the linter default, and go with all spaces as is suggest in this PR, if @rs is OK with it. It's a small replacement job if you copy-paste from Go, but if we get a linter to do the job, it's probably not to much of a hassle.

Linter exception code (be it Markdown Go or otherwise) is a bit unfortunate in my opinion as it hurts readability and maintainability, and ties development to using a given linter software.

That said, I will leave the final word to @rs to decide.

rs:
The linter should not complain about what is in a code blocks. I think it’s a linter problem here.

smyrman:
@SAFAD, would you like to update this PR when you have time to not re-indent the Go code-blocs, and also not add any linter exception comments?

smyrman:
Otherwise I think we will just close it.

SAFAD:
I will be updating this PR during this week, please hold on closing it for now :)

smyrman:
Closing (fixed in new PR)

Issue #182, opened 2018-03-16
schema: Allow FieldSerializer to be called on AnyOff members


smyrman:
### The problem
UPDATE: Let's split out schema.AllOf as a new issue, the solution is to different.

If a `FieldValidator` that implements the `FieldSerializer` interface is wrapped into a `schema.AnyOff` or `schema.AllOff`, the `FieldSerializer` implementation would be ignored, and the field would return it's internal db value instead of the intended serialized one. E.g.

```go
var s = schema.Schema{
    Fields: {
        "myField": {
            Validator: &schema.AnyOf{
                &schema.Null{},
                &CustomValidator{}, // Assumed to implement FieldSerialzer.
            },
        },
    },
}
```
The FieldSerializer implemented by `CustomValidator` would never be detected.

It's very obvious _why_ this does not work since there is no generic way in which the AnyOff or AllOff could determine which FieldSerializer to call (with a possible exception for the AllOff case).

### Possible solution
We might want some time to think of the best solution, but here is at least an initial suggestion.

One possible solution that should cover all cases and not do unnecessary work, would be to let the user _supply_ a serialize "picker" function  (or serialize function, but I will only demonstrate the first) to AnyOff / AllOff when relevant. E.g.:

```go
var s = schema.Schema{
    Fields: {
        "myField": {
            Validator: &schema.AnyOf{
                &schema.Null{},
                &CustomValidator{}, // Assumed to implement FieldSerialzer.
            }.WithSerializePicker(func(value interface{}) int {
                if value == nil {
                    return 0 
                }
                return 1
            }),
        },
    },
}
```

Where `WithSerializePicker` takes a function that returns the index of the FieldValidator that a value is assumed/defined to belong to, and returns a new FieldValidator type, e.g. something like `type WithSerializePicker struct{Validator []FieldValidator, SerializePicker SerializePickerFunc}` (can't be private type because the `jsonschema` package needs to type switch on it).

It is _not_ how ever a solution that _feels_ nice to use.

smyrman:
Probably a nicer solution is to make AnyOf/AllOf a struct.

rs:
For `AnyOff`, why can't we just use the first validator that match as a serializer?

smyrman:
In the example case that wouldn't even necsarliy work, because `CustomValidator` may error on input it does not understand (nil in this case).

Here is another example where it defiantly won't work (some error catch code omitted):

```go
schema.AnyOf{
     &V1{},
     &V2{},
}

[...]

PR #183, opened 2018-03-19
Use first matching serializer when serializing a field with AnyOf

torie:
This PR adds support for calling FieldSerializer on AnyOf members as proposed in #182.

I added a check for the existence of `Serializer` in the test runner to prevent having to add the `Serializer` to every existing test.

### DIFF
--- a/schema/all_test.go
+++ b/schema/all_test.go
@@ -56,6 +56,34 @@ func (tc fieldValidatorTestCase) Run(t *testing.T) {
 	})
 }
 
+type fieldSerializerTestCase struct {
+	Name             string
+	Serializer       schema.FieldSerializer
+	ReferenceChecker schema.ReferenceChecker
+	Input, Expect    interface{}
+	Error            string
+}
+
+func (tc fieldSerializerTestCase) Run(t *testing.T) {
+	t.Run(tc.Name, func(t *testing.T) {
+		t.Parallel()
+
+		if cmp, ok := tc.Serializer.(schema.Compiler); ok {
+			err := cmp.Compile(tc.ReferenceChecker)
+			assert.NoError(t, err, "Validator.Compile(%v)", tc.ReferenceChecker)
+		}
+
+		s, err := tc.Serializer.Serialize(tc.Input)
+		if tc.Error == "" {
+			assert.NoError(t, err, "Serializer.Serialize(%v)", tc.Input)
+			assert.Equal(t, tc.Expect, s, "Serializer.Serialize(%v)", tc.Input)
+		} else {
+			assert.EqualError(t, err, tc.Error, "Serializer.Serialize(%v)", tc.Input)
+			assert.Nil(t, s, "Serializer.Serialize(%v)", tc.Input)
+		}
+	})
+}
+
 type fakeReferenceChecker map[string]struct {
 	IDs       []interface{}
 	Validator schema.FieldValidator
--- a/schema/all_test.go
+++ b/schema/all_test.go
@@ -2,10 +2,10 @@ package schema_test
 
 import (
 	"errors"
+	"reflect"
 	"testing"
 
 	"github.com/rs/rest-layer/schema"
-	"github.com/stretchr/testify/assert"
 )
 
 type referenceCompilerTestCase struct {
@@ -21,9 +21,13 @@ func (tc referenceCompilerTestCase) Run(t *testing.T) {
 
 		err := tc.Compiler.Compile(tc.ReferenceChecker)
 		if tc.Error == "" {
-			assert.NoError(t, err, "Compiler.Compile(%v)", tc.ReferenceChecker)
+			if err != nil {
+				t.Errorf("Compiler.Compile(%v): unexpected error: %v", tc.ReferenceChecker, err)
+			}
 		} else {
-			assert.EqualError(t, err, tc.Error, "Compiler.Compile(%v)", tc.ReferenceChecker)
+			if err == nil || err.Error() != tc.Error {
+				t.Errorf("Compiler.Compile(%v): expected error: %v, got: %v", tc.ReferenceChecker, tc.Error, err)
+			}
 		}
 	})
 }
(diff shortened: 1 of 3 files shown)


schema/anyof.go
```diff
@@ -28,3 +30,22 @@ func (v AnyOf) Validate(value interface{}) (interface{}, error) {
 	// TODO: combine errors.
 	return nil, errors.New("invalid")
 }
+
+// Serialize attempts to serialize the value using the first available
+// FieldSerializer which does not return an error.
```
smyrman: Document fallback

schema/anyof_test.go
```diff
@@ -86,3 +86,32 @@ func TestAnyOfValidate(t *testing.T) {
...
+			Validator:  schema.AnyOf{&schema.Bool{}, &schema.IP{}},
+			Serializer: schema.AnyOf{&schema.Bool{}, &schema.IP{}},
+			Input:      "1.2.3.4",
+			Expect:     "1.2.3.4",
+		},
+		{
+			Name:       `{Bool,IP{StoreBinary:true}}.Serialize("1.2.3.4")`,
+			Validator:  schema.AnyOf{&schema.Bool{}, &schema.IP{}},
+			Serializer: schema.AnyOf{&schema.Bool{}, &schema.IP{StoreBinary: true}},
+			Input:      "1.2.3.4",
```
smyrman: byte array as input

smyrman: Also, must set `StoreBinary` to `true`

smyrman: If possible, implement a dummy Serializer as well so we have a test with two serializers. E.g.

```
Validator: schema.AnyOf{&hexByteArray{},&schema.IP{StoreBinary:true}}
Input: [1, 2, 3, 4]
Output:  "0x1234"
Validator: schema.AnyOf{&schema.IP{StoreBinary:true},& hexByteArray{}}
Input: [1, 2, 3, 4]
Output:  "1.2.3.4"
```

This test would also clearly document the current behaviour in terms of picking a serializer.

We should also have an example with a failing serializer to see that it can pick the right one when the unput types are differen. E.g. 

```
Validator: schema.AnyOf{&intString{},&schema.IP{StoreBinary:true}}
Input: [1, 2, 3, 4]
Output:  "1.2.3.4"
Validator: schema.AnyOf{&schema.IP{StoreBinary:true},&intString{}}
Input: [1, 2, 3, 4]
Output:  "1.2.3.4"
Validator: schema.AnyOf{&intString{},&schema.IP{StoreBinary:true}}
Input: int64(100)
Output:  "100"
Validator: schema.AnyOf{&schema.IP{StoreBinary:true},&intString{}}
Input: int64(100)
Output:  "100"
```

schema/anyof.go
```diff
@@ -2,7 +2,9 @@ package schema
 
 import "errors"
 
-// AnyOf validates if any of the sub field validators validates.
+// AnyOf validates if any of the sub field validators validates. If any of the
+// sub field validators implements the Serialize interface, the *first*
+// serialize function which does not error will be used.
```
smyrman: Serialize -> FieldSerializer
serialize function -> implementation

schema/anyof.go
```diff
@@ -28,3 +30,23 @@ func (v AnyOf) Validate(value interface{}) (interface{}, error) {
 	// TODO: combine errors.
 	return nil, errors.New("invalid")
 }
+
+// Serialize attempts to serialize the value using the first available
+// FieldSerializer which does not return an error. If no appropriate serializer
+// is found, the function returns the input value.
```
smyrman: Make slightly shorter, and avoid calling a method for a function which (although true) can cause slight mental confusion for some: 

the function returns the input value. -> the input value is returned.

smyrman:
@rs this implementation should match what was described as the last solution in #182.

Please comment if you want any changes.

schema/anyof.go
```diff
@@ -28,3 +30,23 @@ func (v AnyOf) Validate(value interface{}) (interface{}, error) {
...
+// FieldSerializer which does not return an error. If no appropriate serializer
+// is found, the input value is returned.
+func (v AnyOf) Serialize(value interface{}) (interface{}, error) {
+	for _, serializer := range v {
+		s, ok := serializer.(FieldSerializer)
+		if !ok {
+			continue
+		}
+
+		v, err := s.Serialize(value)
```
rs: Should we also try the validator first to make sure it's the one that passed?

smyrman: No, I don't think we should. The Validator is supposed to accept the seralized value, e.g. the IP `"1.2.3.4"` as input but it should not be expected to accept the internal value _returened_ from the Validator (e.g. `[]byte{1,2,3,4}` ) as input. That's the Seralize function's job to check (and error on invalid input values), would you not agree?

rs: Make sense.

schema/all_test.go
```diff
@@ -56,6 +56,34 @@ func (tc fieldValidatorTestCase) Run(t *testing.T) {
...
+	Error            string
+}
+
+func (tc fieldSerializerTestCase) Run(t *testing.T) {
+	t.Run(tc.Name, func(t *testing.T) {
+		t.Parallel()
+
+		if cmp, ok := tc.Serializer.(schema.Compiler); ok {
+			err := cmp.Compile(tc.ReferenceChecker)
+			assert.NoError(t, err, "Validator.Compile(%v)", tc.ReferenceChecker)
```
smyrman: Sorry for not pointing this out earlier. I realize this is copy-paste of the existing fieldValidatorTestCase, but we are trying to move away from using testify in Clarify. There is a short discussion about it in #169.

In short, it would be nice (perhaps as a separate commit) to rewrite these tests to standard if checks and calls to `t.Errorf` or `t.Error` on errors.

schema/all_test.go
```diff
@@ -42,16 +46,26 @@ func (tc fieldValidatorTestCase) Run(t *testing.T) {
...
+			if err != nil {
+				t.Errorf("Validator.Validate(%v): unexpected error: %v", tc.ReferenceChecker, err)
+			}
+			if !reflect.DeepEqual(v, tc.Expect) {
+				t.Errorf("Validator.Validate(%v): expected: %v, got: %v", tc.Input, tc.Expect, v)
+			}
 		} else {
-			assert.EqualError(t, err, tc.Error, "Validator.Validate(%v)", tc.Input)
-			assert.Nil(t, v, "Validator.Validate(%v)", tc.Input)
+			if err.Error() != tc.Error {
```
smyrman: may panic if err is nil.

`if err == nil || err.Error() != tc.Error`

schema/all_test.go
```diff
@@ -42,16 +46,26 @@ func (tc fieldValidatorTestCase) Run(t *testing.T) {
...
+				t.Errorf("Validator.Validate(%v): return error: %v, wanted: %v", tc.ReferenceChecker, tc.Error, err)
```
smyrman: swapped order fo got / expected compared to other Errorfs

smyrman: actually the order is the same for tje args, but the text in the format string is turned

schema/all_test.go
```diff
@@ -42,16 +46,64 @@ func (tc fieldValidatorTestCase) Run(t *testing.T) {
...
+		if tc.Error == "" {
+			if err != nil {
+				t.Errorf("Serializer.Serialize(%v): unexpected error: %v", tc.ReferenceChecker, err)
+			}
+			if reflect.DeepEqual(s, tc.Expect) {
+				t.Errorf("Serializer.Serialize(%v): expected: %v, got: %v", tc.Input, tc.Expect, s)
+			}
+		} else {
+			if err.Error() != tc.Error {
+				t.Errorf("Serializer.Serialize(%v): return error: %v, wanted: %v", tc.ReferenceChecker, tc.Error, err)
```
smyrman: Same here:
`expected error: %v, got: %v`

schema/all_test.go
```diff
@@ -21,9 +21,13 @@ func (tc referenceCompilerTestCase) Run(t *testing.T) {
...
 		err := tc.Compiler.Compile(tc.ReferenceChecker)
 		if tc.Error == "" {
-			assert.NoError(t, err, "Compiler.Compile(%v)", tc.ReferenceChecker)
+			if err != nil {
+				t.Errorf("Compiler.Compile(%v): unexpected error: %v", tc.ReferenceChecker, err)
+			}
 		} else {
-			assert.EqualError(t, err, tc.Error, "Compiler.Compile(%v)", tc.ReferenceChecker)
+			if err == nil || err.Error() != tc.Error {
+				t.Errorf("Compiler.Compile(%v): return error: %v, wanted: %v", tc.ReferenceChecker, tc.Error, err)
```
smyrman: expected error: %v, got: %v"

PR #186, opened 2018-04-10
rest list DELETE request now parse filter and window correctly

### LINKED ISSUE #127: Clear does not support filters or limit

smyrman:
Allow query parameters (limit, skip, page, sort and filter) to take effect on (list) DELETE requests in the `rest` package. This resolves #127, although an update to the rest-layer-mongo Storer backend (and potentially others) is required for pagination to take effect when that backend is used. This is tracked by rs/rest-layer-mongo#25.

TODO:
- [x] Fix the issue in the route parser
- [x] ~Merge fix for rest-layer-mem or~ move rest-layer-mem to rest-layer repo (out of vendor).
- [x] Update tests in affected (failing) test-files to use the external interface (HTTP) for testing.
- [x] Manualy tested with rs/rest-layer-mongo#26 (sanity check)


smyrman:
@torie, do you want to do an initial review on this?

README.md
```diff
@@ -1042,15 +1042,13 @@ Such request can quickly generate a lot of queries on the storage handler. To en
...
+Pagination is supported on collection URLs using the `page` and `limit` query-string parameters, and can be used for resource list view URLs with request method `GET` and `DELETE`. If you don't define a default pagination limit using `PaginationDefaultLimit` resource configuration parameter, the resource won't be paginated for list `GET` requests until you provide the `limit` query-string parameter. The `PaginationDefaultLimit` does not apply to list `DELETE` requests, but the `limit` and `page` parameters may still be used to delete a subset of items.
 
 If your collections are large enough, failing to define a reasonable `PaginationDefaultLimit` parameter may quickly render your API unusable.
 
 ## Skipping
 
-Skipping of resource items is defined through the `skip` query-string parameter. The `skip` value is a positive integer defining the number of items to skip when querying for items.
-
-To use a resource field with the `skip` parameter, the field must be defined on the resource.
```
smyrman: > To use a resource field with the `skip` parameter, the field must be defined on the resource.

@rs, I was confused by the meaning of this, and I have removed it for now. What is it trying ti say? What is resource field in this context? A schema field? If we want to add it back, a corresponding example may be appropriate (at least for me), so that it can e better understood.

schema/query/sort.go
```diff
@@ -50,16 +50,16 @@ func ParseSort(sort string) (Sort, error) {
...
-		if field == nil {
-			return fmt.Errorf("invalid sort field: %s", sf.Name)
+		f := validator.GetField(sf.Name)
+		if f == nil {
+			return fmt.Errorf("%s: unknown sort field", sf.Name)
 		}
-		if !field.Sortable {
-			return fmt.Errorf("field is not sortable: %s", sf.Name)
+		if !f.Sortable {
+			return fmt.Errorf("%s: field is not sortable", sf.Name)
```
smyrman: FYI, this is changed to make the error messages for the `sort` and `filter` query-string parameters consistent.

README.md
rs: I think it is no longer relevant.

smyrman:
FYI: I have been waiting with merging this until I can find some time to implement (rs/rest-layer-mongo#25) and do a sanity check with that backend. If I can't find that time before the end of this weekend, I will just merge this changes; they will still work with the mongo backend, but there will be a _bug_ when using the window parameters.

rs:
Sounds good.

rest/method_post_test.go
```diff
@@ -148,7 +151,7 @@ func TestHandlerPostList(t *testing.T) {
 				return http.NewRequest("POST", "/test", bytes.NewBufferString(`{}`))
 			},
 			// FIXME: The HTTP 520 code is usually used for protocol errors, and
-			// seems unaprporiate. This should most likely be a 422 error.
+			// seems unappropriated. This should most likely be a 422 error.
```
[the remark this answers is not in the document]
smyrman: fixed

Issue #187, opened 2018-05-25
Behaviour of query/predicate does not match behaviour of MongoDB storage backend


torie:
The behaviour of the filter operators in the `query` package does not work when the document field being matched against is an array. For example, given the following document and query, the MongoDB storage backend will return a match, while the `resource/testing/mem` backend will not.

### Document
`
  {
    "foo": ["bar", "baz"]
  }
`

### Query
`
{ field: { $in: ["bar"] } }
`

This is the case for the following operators, `$eq`, `$in` `$ne`, `$nin`.

smyrman:
Support for `$in` and `$nin` added in #188.
Support for `$eq` and `$ne` still need to be updated.

smyrman:
https://docs.mongodb.com/manual/reference/operator/query/eq/#array-element-equals-a-value

smyrman:
Accidentally clicked the close button...

Dragomir-Ivanov:
I guess I can take on this.

Dragomir-Ivanov:
Hmm, that is strange. First there is no `$eq` operator in `rest-layer`. I added these tests in `predicate_test.go` and they are seem to be passing. I will test with real resource tomorrow, though.

```
		{
			`{"foo":["bar","baz"]}`, []test{
				{map[string]interface{}{"foo": []interface{}{"bar", "baz"}}, true},
				{map[string]interface{}{"foo": []interface{}{"bar"}}, false},
				{map[string]interface{}{"foo": []interface{}{"baz"}}, false},
				{map[string]interface{}{"foo": []interface{}{"bar", "baz", "tar"}}, false},
				{map[string]interface{}{"foo": []interface{}{}}, false},
			},
			nil,
		},
		{
			`{"foo":{$ne: ["bar","baz"]}}`, []test{
				{map[string]interface{}{"foo": []interface{}{"bar", "baz"}}, false},
				{map[string]interface{}{"foo": []interface{}{"bar"}}, true},
				{map[string]interface{}{"foo": []interface{}{"baz"}}, true},
				{map[string]interface{}{"foo": []interface{}{"bar", "baz", "tar"}}, true},
				{map[string]interface{}{"foo": []interface{}{}}, true},
			},
			nil,
		},
```

Dragomir-Ivanov:
@torie Any comment appreciated.

smyrman:
```
{
			`{"foo":"bar"}`, []test{
[...]

PR #188, opened 2018-05-25
Predicate operators

torie:
This add support for using the $in and $nin operators in the query package against array fields as described in https://github.com/rs/rest-layer/issues/187.

It also adds tests for nested fields and the `utils.getField` function.


smyrman:
@rs, this allows the `mem` backend to work more equal to how MongoDB works for `$in` and `$nin` queries, making it easier to write unit-tests when using that backend. I don't have any experience with the other rest-layer backends, such as e.g. elastic-search https://github.com/rs/rest-layer-es/blob/c9cca01e407ccc54e6f84f228ac0c45700c5cf96/query.go#L81.

PR #191, opened 2018-06-16
gofmt

whilei:
Just ran

	gofmt -w .

on the project root. That's all.

https://blog.golang.org/go-fmt-your-code

---
> I made this PR with a project going on over at https://github.com/rotblauer/gofmt-att, and it's definitely a work in progress. So if I got something wrong, or this is annoying at all, please file an issue over there and we'll sort it out.

### DIFF
--- a/rest/resource_path_test.go
+++ b/rest/resource_path_test.go
@@ -3,8 +3,8 @@ package rest
 import (
 	"testing"
 
-	mem "github.com/rs/rest-layer/resource/testing/mem"
 	"github.com/rs/rest-layer/resource"
+	mem "github.com/rs/rest-layer/resource/testing/mem"
 	"github.com/rs/rest-layer/schema"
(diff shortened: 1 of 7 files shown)


rest/resource_path_test.go
```diff
@@ -3,8 +3,8 @@ package rest
 import (
 	"testing"
 
-	mem "github.com/rs/rest-layer/resource/testing/mem"
 	"github.com/rs/rest-layer/resource"
+	mem "github.com/rs/rest-layer/resource/testing/mem"
```
smyrman: Could remove `mem ` prefix for consistency - not needed, and no longer different from the last bit of the import path

whilei: @smyrman You're right... but for me right now that's an "out of scope" decision for this PR because it's not a strict change by `gofmt`. Please feel welcome to make modifications however you want, though.

PR #196, opened 2018-09-16
rest: avoid nil map assignment of url.Values

### LINKED ISSUE #195: rest: asing to nil map
When using a resource query Alias, I got a panic in the rest package due to an assignment to nil map at the following line:

https://github.com/rs/rest-layer/blob/48c11e50091e8c988780853a3aaa8af25cf98d42/rest/routing.go#L126 (the panic happens inside go's url.Values.Add).

Looks like the url.Values isn't necessarily initialized...

smyrman:
Fixes #195 

---
TODO:

- [x] go get is failing on travis; figgure out why.


smyrman:
@rs, do you want to re-approve?

Issue #170, opened 2017-11-22
Improve query operator validator


smyrman:
As discussed in #149 (after the fix), query operators (`$ne`, `$regex`, `$lt` etc.)  currently get their values validated with the associated `FieldValidator` as normal values which is inaccurate.

 E.g. ` ?filter={name:{$regex:"^test.*$"}}`, `"^test.*$"` would be validated by the `FieldValidator` for `name`, which is inaccurate in the case where `name` has additional validation criteria, such as `Allowed: []string{"testA", "testV"}` or `Regex`.

What we instead want to validate is,
- Does the operator make sense for the assicated `FieldValidator`?
- Does the operator value have the correct type according to the selected `FieldValidator`?

One way to do this, could be a new optional interface for `FieldValidator` as mentioned in #149.

smyrman:
Interface suggestion:

```go
// OpValidator is a an optional interface for FieldValidator implementations use to allow 
// operator queries when using the rest-layer schema/query package.
type OpValidator interface{
    // ValidateOperator should validates a value operator such as $ne, $lt, $gt, $regex etc, and return
    // either a value that can be used for querying the storage backend, or an error if the operator is
    // not allowed, or the operator value is invalid.
    ValidateOperator(op string, value interface{}) (interface{}, error)
}
```

Dragomir-Ivanov:
@smyrman Can you give one more complete example why we need this. These operators are query only, so they don't update any values, thus is `Allowed` even applicable to them?
Maybe there is a good case for your proposal, I just don't completely understand.

smyrman:
> thus is Allowed even applicable to them?

That is exactly the point, it is not, so when validating an operator values, additional options set in the `FieldValidator` should be ignored.

E.g. `name:{$regex:"^test.*$"} would fail to validate (although it's a valid query), if the `FieldValidator` of `name` (most likely a `schema.String`), provides additional validation options such as `Allowed`, `MinLen`, MaxLen` or `Regex`.  Therefore we *can not* simply call `Validate` on the field's `FieldValidator` instance.

The same is true for numbers. `number:{$lt:1000}` is valid even if the `FieldValidator` for `number` has an upper limit for the value set to `900`.

In adition to that, I suspec that adding the new optional interface could allow even better queries, as we might be able to use the interface e.g. to allow `$lt` and `$gt` queries on non-numeric fields such as e.g. `schema.Time`.

Dragomir-Ivanov:
[...]

PR #204, opened 2018-10-20
Implement predicate type conversion.

Dragomir-Ivanov:
Helps with querying for `schema.Time` and others. It also skips validation for `schema.*` types during predicate phase of query.
This is just a preview PR, I will fix the tests, when implementation is accepted.


schema/field.go
```diff
@@ -94,17 +94,19 @@ type FieldHandler func(ctx context.Context, value interface{}, params map[string
 // value to validate as argument and returned the normalized value or an error
 // if validation failed.
 type FieldValidator interface {
-	Validate(value interface{}) (interface{}, error)
+	// Validate a field with a value. With enforce=false this function only converts to internal type.
+	// enforce=true activates all other per type checks
+	Validate(value interface{}, enforce bool) (interface{}, error)
```
smyrman: This really isn't the right interface to allow for query operator validation.
1. It does two different things based on a Boolean value -- design-wise this means it should be two methods, not one, at the very least at the public interface level.
2. It doesn't really handle all the cases it needs to handle for operator value validation; It _doesn't_ only need to validate that the value is correct, it also needs to check that the operator itself makes sense for the given type. E.g. `$regex` don't make sense for time or numbers, and `$gt` does not make sense for strings.

This is why the suggested interface in #170 includes an optional interface that takes both the operator and the value as input parameters. Perhaps strings for operators isn't the best choice though, and maybe simialr operations can be grouped. E.g. if `$eq` is allowed, `$ne` should be allowed, if `$gt` is allowed, `$lt` should be allowed"

```go
type QueryOp uint

const (
    QueryOpEquality = iota // allow $eq, $ne
    QueryOpLessGreater // allow $lt, $eq
    QueryOpIn // allow $in $nin
    QueryOpRegex // allow $regex
)

type OperatorValidator interface {
    // Return an error if op isn't allowed, or if value isn't of an acceptable type or format.
    ValidateOperator(op QueryOp, value interface{}) (interface{}, error)
}
```

Dragomir-Ivanov: Hmm, I just did the following with the patch above:
```
GET http://localhost:3001/api/visits?filter={name: {$regex: 5 }}
```
I am getting this result:
```
{
  "code": 422,
  "issues": {
    "filter": [
      "char 16: name: $regex: not a string"
    ]
  },
  "message": "URL parameters contain error(s)"
}
```

So I think matching types are covered. Also Since Mongo allows using string lexical comparison, why we don't allow it as well? 
Let me try polish some more.

schema/query/predicate.go
```diff
@@ -303,97 +333,130 @@ func (e NotExist) Validate(validator schema.Validator) error {
...
-	Value float64
+	Value Value
 }
 
 // Match implements Expression interface.
 func (e LowerOrEqual) Match(payload map[string]interface{}) bool {
-	n, ok := isNumber(getField(payload, e.Field))
-	return ok && (n <= e.Value)
+	// n, ok := isNumber(getField(payload, e.Field))
+	// return ok && (n <= e.Value)
```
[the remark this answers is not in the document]
Dragomir-Ivanov: Yes, I guess there is a lot of work for `rest-layer-mem` to support all MongoDB functionality and pecularities. For now my intention was to just hard-code support for int/float/Time into `Match` function. But I can try your proposal here.

Dragomir-Ivanov:
@smyrman My initial concern is:
1. The name for the validation function `ValidateOperator`, is not quite right, since we are converting the current `Expression` and return a new one. So I guess more appropriate name is needed to reflect that.
2. `Predicate is defined as `[]Expression` but we still need to comply with `Expression` interface, so `ValidateOperator` is just a noise for `Predicate` type.

I do like however the way whole operator validation step is separate from other steps, and can be opted-out if needed.

smyrman:
Why is `ValidateOperator` needed at the `Expression` and`Predicate` level? What does it do? Isn't it enough with the new FieldValidator method?

smyrman:
It looks like it validates that the operator is allowed? Can't that be done as part of the normal Validate?

Dragomir-Ivanov:
> It looks like it validates that the operator is allowed? Can't that be done as part of the normal Validate?

I am not sure I understand. You want `FieldValidator` `Validate(value interface{}) (interface{}, error)` to accept the `op` as well, and make the check on the spot? But I think `Validate` is used on other places apart from Query parsing.

Dragomir-Ivanov:
> > Why is `ValidateOperator` needed at the `Expression` and`Predicate` level? What does it do? Isn't it enough with the new FieldValidator method?
> 
> I think your proposal was separate interface `OperatorValidator` and not adding new method to `FieldValidator` however I am all for the latter case, and will make the change if you say so.

Also I will try to remove `ValidateOperator` function from `Expression` interface and use a global function.

Dragomir-Ivanov:
@smyrman Are we calling interface functions, and struct functions `method` in Go? I thought `method` was for OOP principles?

smyrman:
> I am not sure I understand. You want FieldValidator Validate(value interface{}) (interface{}, error) to accept the op as well, 

no. I mean _why_ does the Predicate and Expression need a new method:

```go
// ValidateOperator TODO
func (e Predicate) ValidateOperator(validator schema.Validator) (Expression, error) {
	return nil, validateOperatorExpressions(e, validator)
}

// Expression is a query or query component that can be matched against a payload.
type Expression interface {
	Match(payload map[string]interface{}) bool
	Validate(validator schema.Validator) error
	ValidateOperator(validator schema.Validator) (Expression, error)
	String() string
}
```


> I think your proposal was separate interface OperatorValidator and not adding new method to FieldValidator however I am all for the latter case, and will make the change if you say so.

yes, that was my proposal (to add it as a new interface), and I think that is still the best approach for now.

smyrman:
> @smyrman Are we calling interface functions, and struct functions method in Go? I thought method was for OOP principles?

The concept, in my mind, is the same regarding of language.

> Quote: A method is a function with a special receiver argument. 

- https://tour.golang.org/methods/1

schema/query/predicate.go
```diff
@@ -216,12 +222,18 @@ func (e NotIn) String() string {
...
 // Validate implements Expression interface.
 func (e NotIn) Validate(validator schema.Validator) error {
-	return validateValues(e.Field, e.Values, validator)
+	return validateValues(e.Field, e.Values, schema.OpNotIn, validator)
+}
+
+// Fix TODO
+func (e NotIn) Fix(validator schema.Validator) (Expression, error) {
+	newValues, err := validateValuesOperator(e.Field, e.Values, schema.OpNotIn, validator)
+	return NotIn{Field: e.Field, Values: newValues}, err
```
smyrman: It looks to me like you should set the "newValues" during the initial ParsePredicate -- is there a reason that won't work?

smyrman: Right, I see the problem now. I will try to find some time and look at the code a bit deeper, and then get back to you. I suppose it _could_ maybe be done as part of Validate if it's changed to use pointer receivers all over the place...

I will try to find some time over the weekend.

schema/query/predicate.go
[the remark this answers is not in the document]
Dragomir-Ivanov: For the `a == b` and `a != b` case I guess we have more generic implementation using `reflect.DeepEqual`. Shouldn't we keep it. Also if a type implements `Comparable` then it would support `$gt $gte $lt $lte` implicitly so there is no need to restate that in `ValidateOprator` method for each type.

schema/query/predicate.go
smyrman: Thanks for your work!

schema/query/predicate.go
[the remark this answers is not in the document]
Dragomir-Ivanov: @smyrman Any comment on the above?

schema/query/predicate.go
smyrman: I started looking at where `ParsePredicate` is being called, and that gives away waht's availabe at these locations in the code -- e.g. is the schema validator available?

These are the non-test results I found

https://github.com/rs/rest-layer/blob/48c11e50091e8c988780853a3aaa8af25cf98d42/schema/query/query.go#L69

https://github.com/rs/rest-layer/blob/9197c24a34866e9717ca55c4364df76c362d95dc/rest/routing.go#L261

https://github.com/rs/rest-layer/blob/48c11e50091e8c988780853a3aaa8af25cf98d42/graphql/query.go#L114

https://github.com/rs/rest-layer/blob/48c11e50091e8c988780853a3aaa8af25cf98d42/graphql/query.go#L125


The first case is query.New, which is a bit of a special case, and I can't find it used outside of tests.

The next three usages all have the `schema.Validator` interface available, and in fact, the `.Validate` method is called on each instance here.

I am thinking all of the following are possible solutions that would work, and we can discuss the design:

1. Validate is changed to use pointer receivers, and modifes the parent predicate/expression.
2. Validate is set to return a new predicate -- it can optionally be ignored when it's not needed.
3. A second inializer, ParsePredicateForValidator(schema.Validator) is introduced.

In the case of design, I feel 2 is the best fit for rest-layer for the following reasons:
- The mutation is explicit.
- The interface is kept similar to the interface of schema.Validate, that also returns validated type and an error.
- All the other query components (ordering and projection) calls .Validate -- continuing to call .Validate is the least amont of change that would allow for a, at least mostly, consistent API.

@rs -- If you have a view on the design of this, it would be great to know.

The problem we want to solve, is more or less the one described in #170. I believe @Dragomir-Ivanov is particular keen on solving it for the case of `$lt` and `$gt` for time.Time, and potentially a custom `Deciaml` type. Solving it in a generic way would be nice, so that it's compatible with third-party FieldValidators and database storage types.

schema/query/predicate.go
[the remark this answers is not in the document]
smyrman: Hi @Dragomir-Ivanov . That may be a good approach I suppose -- it doens't solve the case of regex, but perhaps that should be solved separately?

On the schema level, would it then be that a FieldValidator implementation that is to support `$lt[e]`/`$gt[e]` implements an interface ala:

```go
type ComparableField struct {
    /// ComparableValue returns a comparable and true, or nil and false.
    ComparableValue(value interface{}) (Comparable, bool)
}
```

Technically, it doesn't need the "ok" `bool` return, it could just return `nil` when it's not possible to convert to the target value, but keeping the second parameter kind of makes the intent more clear, and that it can return nil.

Same reasoning goes for the `AsComparable`. I don't know what's the best design, but as long as the behaviour is documented, I would accept either.

And maybe `string` types that supports Regex search could implement (different PR):

```go
type RegexField struct {
    CompareableToRegex() bool
}
```
```go
type Comparable interface {
	// AsComparable converts other to a comparable type if possible.
	AsComparable(other interface{}) (Comparable, bool)

	// Less returns true if other is comparable and less than this value.
	Less(other interface{}) bool
}
```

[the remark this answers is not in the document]
smyrman: It doesn't change the need to pass a schema.Validator to the `query.Predicate` in order to get the comparable value, but it may be a bit more clear in the `schema` pacakge, and we won't need to move or expose all the operator value definitions.

schema/query/predicate.go
smyrman: I intentionally skipped the MustParsePredicate.

It's all non-tests inside _rest-layer_ where it's called. `MustParsePredicate` would probablly be called in end-consumer applications, and would need to stay either way, as it's recommended for use in `schema` definitions etc.

Dragomir-Ivanov: Yes, but how we are going to proceed in this case. There is no `schema.Validator` declaration in this context? `ParsePredicate` may accept it as a pointer value, so we can pass `nil` in case whether it is not accessible. However I feel that we need always have the schema validator where we are making a query.

smyrman: > ParsePredicate may accept it as a pointer value, so we can pass nil in case whether it is not accessible. However I feel that we need always have the schema validator where we are making a query.

It's mostly available, but there seem to be two exceptions.

1. When you defined a Field Dependency - This is the main use-case for `MustParsePredicate`, the schema isn't ready yet, so no schema.Validator can be passed in at parse-time.
2. The internal code for evaluating a projection on a connected field (the last `ParsePredicate` code reference I previously overlooked),  don't have access to the related schema today, and it does not look like it even calls validate on the parsed predicate before using it.

Case 1, from what I can see, get's validated later:
- https://github.com/rs/rest-layer/blob/48c11e50091e8c988780853a3aaa8af25cf98d42/schema/dependency.go#L23

Case 2 doesn't seam to be validated at all today (unless I somehow missed it), which seams like a bug.

smyrman: Check that it's the right one though... you want the schema.Validator of the connected schema, not the schema you are _at_.

Connections are used to _embed_ other resources.

smyrman: https://github.com/rs/rest-layer#embedding

Dragomir-Ivanov:
Hi @smyrman @rs , here is my design proposal for this PR. 
Sindre, I didn't quite followed your design on `Comparabe` interface, but I think this will still be OK.

Lets define 2 new interfaces that can be implemented by `schema.FieldValidator`
```
type FieldComparator interface {
	Less(this interface{}, other interface{}) bool
}

type FieldPreparator interface {
        // This can be named Parse as well. I would like to go into schema.FieldValidator interface,
        // so it will be mandatory for all schema types.
	Prepare(value interface{}) (interface{}, error)
}
```

`FieldComarator` will be used in the `Expression.Match` method, and can give custom types `rest-layer-mem` support. Perfect candidates will be `schema.Integer, schema.Float, schema.Time` and possibly others.
`FieldPreparator` will be used in `ParsePredicate` in order to make sure all query types match the schema defined types. This will not do full type `Validate` but will only make sure the types are compatible. It will still put as `Validate` the type in the `Expression` expected by the DB/Storer. 

In order to accomplish the above, I needed to refactor a little bit passed arguments of some functions involved in predicate parsing. I will not go into much details here, because code will explain better.
No change in the public API is made. Please when reviewing this PC, use the `Files Changed` tab in the GitHub, because last commit undoes much of the previous work.
Thanks for the help, guys!

rs:
Looks good to me.

schema/dependency.go
```diff
@@ -39,8 +39,9 @@ func (s Schema) validateDependencies(changes map[string]interface{}, doc map[str
 		path := prefix + name
 		field := s.GetField(path)
 		if field != nil && field.Dependency != nil {
-			if !field.Dependency.Match(doc) {
-				addFieldError(errs, name, fmt.Sprintf("does not match dependency: %+v", field.Dependency))
+			p := field.Dependency(s)
```
Dragomir-Ivanov: I guess I need to precache that `Predicate` somewhere, and not parse it every time.

schema/integer.go
```diff
@@ -12,8 +12,8 @@ type Integer struct {
 	Boundaries *Boundaries
 }
 
-// Validate validates and normalize integer based value.
-func (v Integer) Validate(value interface{}) (interface{}, error) {
+// Prepare TODO
+func (v Integer) Prepare(value interface{}) (interface{}, error) {
 	if f, ok := value.(float64); ok {
 		// JSON unmarshaling treat all numbers as float64, try to convert it to
 		// int if not fraction.
```
Dragomir-Ivanov: `I do think that in the `Prepare` phase, we should allow `float` values to be converted to `integer`. 
I.e. `?filter={someint: {$gt: 23.44}}` should be valid. However accepting `float` in `PUT,POST,PATCH` should be forbidden.

schema/query/predicate_validator.go
```diff
@@ -15,6 +15,17 @@ func validateExpressions(exps []Expression, validator schema.Validator) error {
 	return nil
 }
 
+func prepareExpressions(exps []Expression, validator schema.Validator) ([]Expression, error) {
+	for i, exp := range exps {
+		if nexp, err := exp.Prepare(validator); err != nil {
+			return nil, err
+		} else {
+			exps[i] = nexp[0]
```
Dragomir-Ivanov: Is this the right way to do it? Does `exps` can be reused, and not defining `newExpressions` variable?

schema/query/predicate_validator.go
```diff
@@ -52,29 +63,46 @@ func validateValue(field string, value Value, validator schema.Validator) error
...
-			}
-		default:
-			return fmt.Errorf("%s: cannot apply %s operation on a non numerical field", field, op)
+		if preparator, ok := f.Validator.(schema.FieldPreparator); ok {
+			for i, value := range values {
+				nv, err := preparator.Prepare(value)
+				if err != nil {
+					return nil, err
 		}
+				values[i] = nv
```
Dragomir-Ivanov: Same, can we reuse `values` here, and not defining `newValues`?

smyrman: no, need to define new values.

schema/query/predicate_validator.go
smyrman: actually, `[]Expression` also implements `Expression`... so that should be a valid return value as well, and you don't need the index `0` or `...` in append.

Dragomir-Ivanov:
@smyrman Committed with your suggestions. Didn't implemented Integer field with Float compare value. Lets leave that for later, when we clean things up. 
Wanted to add `ValidateQuery` method to `schema.FieldValidator` however, `schema.Schema` also needs to implement method for `FieldValidator`, and I think that is wrong. We can discuss however. 
I did some simple tests, it seems all to work, however when I refactor all broken tests we can have clear picture.

smyrman:
Think this approach looks good @Dragomir-Ivanov 👍 Looking forward to the result :-) I think it's fair not to add `ValdateQuery` to the `FieldValidator` inteface for now.

Dragomir-Ivanov:
Thanks @smyrman , can you take quick look on the questions above?

schema/query/predicate.go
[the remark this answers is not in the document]
smyrman: Closing this conversation, as there is now a better proposal for the interface.

schema/dependency.go
smyrman: If you leave it as an "unprepared" query instead of a func, the `resource.Index.Bind` and `resource.Resource.Bind` methods could potentially prepare and store it. It does a similar form of override where it creates `schema.Connection` fields (the reverse of `schema.Resource`)

Dragomir-Ivanov:
I actually made a couple of code comments above on the last commit, since I am not sure how to best Go approach.

smyrman:
> I actually made a couple of code comments above on the last commit, since I am not sure how to best Go approach.

Can't see any comments in neither Changed Files nor the last commit itself -- sure they are not with state pending? If they are you need to "submit a review" for me to see them.

schema/query/predicate.go
```diff
@@ -52,15 +52,15 @@ func (e Predicate) String() string {
 	return "{" + strings.Join(s, ", ") + "}"
 }
 
-// Validate implements Expression interface.
-func (e Predicate) Validate(validator schema.Validator) error {
-	return validateExpressions(e, validator)
+// Prepare implements Expression interface.
+func (e *Predicate) Prepare(validator schema.Validator) error {
```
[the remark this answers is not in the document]
smyrman: No, only this one, and technically, not for slice values. Predicate would _work_ without a pointer value as long as it doesn't change the _size_ of the slice. That said, using a pointer may make the intention more clear...

smyrman:
PS! Can you please add a note to the "breaking changes" section of the README with a link to this PR?

Breaking changes is defined as anything that changes the public interface in a way that would cause old code to no longer compile.

smyrman:
Before we do the final merge, please also do a final rebase and squah and write a nice commit message :-)

schema/query/predicate_validator.go
```diff
@@ -31,50 +31,62 @@ func validateField(field string, validator schema.Validator) error {
...
+		// if there is explicit validator for quering
+		validateFunc = qv.ValidateQuery
+	}
+		for i, v := range values {
+		nv, err := validateFunc(v)
+		if err != nil {
 				return fmt.Errorf("%s: invalid query expression `%#v': %v", field, v, err)
-			}
-		}
+				}
```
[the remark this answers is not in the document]
Dragomir-Ivanov: I am using VSCode, with the Golang plugin, so I have these enabled. Don't know what happened.

schema/query/predicate_validator.go
```diff
@@ -31,50 +31,62 @@ func validateField(field string, validator schema.Validator) error {
...
-			if _, err := f.Validator.Validate(value); err != nil {
-				return fmt.Errorf("%s: invalid query expression: %v", field, err)
-			}
-		default:
-			return fmt.Errorf("%s: cannot apply %s operation on a non numerical field", field, op)
+		return nil, err
 		}
+	c, ok := f.Validator.(schema.FieldComparator)
+	if !ok {
+		// XXX. Should we return error, stating that field is non-comparable?
```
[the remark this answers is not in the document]
Dragomir-Ivanov: Actually this will complain that it is not supported for `rest-layer-mem`, since other backends don't use `Match` for `$gt $gte $lt $lte`. But I think this is a good policy to `warn/error` in this case.

Dragomir-Ivanov:
@smyrman This is what I recall for the breaking changes part. Do you recall something else?
```
PR #204:
Storage drivers need to accept pointer to Expression implementer in query.Predicate.
filter parameters in sub-query will be validated for type match.
filter parameters will be validated for type match only, instead of type & constrains.
```

schema/float_test.go
```diff
@@ -7,6 +7,27 @@ import (
...
+	assert.Nil(t, s)
+	s, err = Float{Boundaries: &Boundaries{Min: 0, Max: 2}}.ValidateQuery(3.1)
+	assert.NoError(t, err)
+	assert.Equal(t, 3.1, s)
+	s, err = Float{Boundaries: &Boundaries{Min: 0, Max: 2}}.ValidateQuery(1.1)
+	assert.NoError(t, err)
+	assert.Equal(t, 1.1, s)
+	s, err = Float{Boundaries: &Boundaries{Min: 2, Max: 10}}.ValidateQuery(1.1)
+	assert.NoError(t, err)
+	assert.Equal(t, 1.1, s)
```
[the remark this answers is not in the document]
Dragomir-Ivanov: What about the rest of the tests in the test suit? Wouldn't be awkward to have part of the tests `assert` free, and other part full of asserts? I guess you want me to redo just new tests I have added.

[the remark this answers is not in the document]
smyrman: If you redo your new tests, that's more than good enough for me. It will be a bit acward to have two different styles within the same file -- could perhaps be fixed by simply putting the new tests in a new file.

- <name>_test.go <- new tests
- <name>_assert_test.go <- rename file with old tests...

schema/query/predicate_validator.go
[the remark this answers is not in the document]
smyrman: Let's return an error then.

schema/float_test.go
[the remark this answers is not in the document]
Dragomir-Ivanov: Okay, I will redo my tests only for now. Lets leave separating the test files for another PR. Depending on how much work it is, I might just redo the old tests as well.

Dragomir-Ivanov:
@smyrman Pushed the squashed rebase. When you merge it, I will do additional PRs to fix the DB drivers.

smyrman:
Yes. Third-party drivers seam out of date.

schema/array.go
```diff
@@ -26,21 +26,51 @@ func (v *Array) Compile(rc ReferenceChecker) (err error) {
...
+	}
+	return val, nil
+}
+
+// Validate implements FieldValidator.
+func (v Array) Validate(value interface{}) (interface{}, error) {
+	arr, err := v.validate(value, false)
+	if err != nil {
+		return nil, err
+	}
```
smyrman: Performance wise, it seams more sane to do the lengh check first. That probably means doing the type assertion to an array before calling `validateValues`.

smyrman:
Sorry for comming with even more comments now... I sort of look at this in between other things and haven't been able to really spot everything in my first pass through. Thanks a lot for your great work!

Dragomir-Ivanov:
No problem Sindre, I am learning a lot of things during these passes, so let it be as many as needed :)
I have made a new commit addressing your suggestions. Did separate commit so it will be easier for you to review. Then you can make q `squash merge` or I can squash the PR if you want.

smyrman:
@Dragomir-Ivanov can you squash the last fixup commits as well? If there are no more comments, I think this is good for merging and we can deal with any fallout if there is anything we haven't spotted yet.

Dragomir-Ivanov:
@smyrman Done.

PR #205, opened 2018-11-26
Fix rest-layer-mem behaviour regarding array fields to match MongoDB.…

### LINKED ISSUE #187: Behaviour of query/predicate does not match behaviour of MongoDB storage backend

Dragomir-Ivanov:
… Fixes #187

### DIFF
--- a/schema/query/predicate.go
+++ b/schema/query/predicate.go
@@ -232,7 +232,23 @@ type Equal struct {
 
 // Match implements Expression interface.
 func (e Equal) Match(payload map[string]interface{}) bool {
-	return reflect.DeepEqual(getField(payload, e.Field), e.Value)
+	value := getField(payload, e.Field)
+
+	// The equality operator in MongoDB allows matching both single values and an
+	// array of values.
+	// https://docs.mongodb.com/manual/tutorial/query-arrays/#query-an-array-for-an-element
+	vt, vok := value.([]interface{})
+	_, eok := e.Value.([]interface{})
+	if vok && !eok {
+		for _, vv := range vt {
+			if reflect.DeepEqual(e.Value, vv) {
+				return true
+			}
+		}
+	} else {
+		return reflect.DeepEqual(value, e.Value)
+	}
+	return false
 }
 
 // Prepare implements Expression interface.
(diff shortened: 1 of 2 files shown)


schema/query/predicate.go
```diff
@@ -232,7 +232,23 @@ type Equal struct {
 
 // Match implements Expression interface.
 func (e Equal) Match(payload map[string]interface{}) bool {
-	return reflect.DeepEqual(getField(payload, e.Field), e.Value)
+	value := getField(payload, e.Field)
+
+	// The $eq(not present in rest-layer) operator in MongoDB allows matching both single values and an
```
smyrman: I guess you can state "the equality operator" instead of "$eq". What was meant was the implicit equality check as represented by `query.Equal`. Writing `$eq` in the issue was inaccurate.

schema/query/predicate.go
```diff
@@ -255,7 +271,23 @@ type NotEqual struct {
 
 // Match implements Expression interface.
 func (e NotEqual) Match(payload map[string]interface{}) bool {
-	return !reflect.DeepEqual(getField(payload, e.Field), e.Value)
+	value := getField(payload, e.Field)
+
+	// The $ne operator in MongoDB allows matching both single values and an
+	// array of values.
```
smyrman: FYI: Just did a small test in the MongoDB shell to confirm this is how `$ne` works.

```
> db.c.insert({"a":["a","b"]})
WriteResult({ "nInserted" : 1 })
> db.c.insert({"a":["c","b"]})
WriteResult({ "nInserted" : 1 })
> db.c.insert({"b":"z"})
WriteResult({ "nInserted" : 1 })
> db.c.find({"a":"a"})
{ "_id" : ObjectId("5bfc84458014a3148be6f734"), "a" : [ "a", "b" ] }
> db.c.find({"a":{"$ne":"a"}})
{ "_id" : ObjectId("5bfc844f8014a3148be6f735"), "a" : [ "c", "b" ] }
{ "_id" : ObjectId("5bfc846e8014a3148be6f736"), "b" : "z" }
```

PR #206, opened 2018-11-27
Prevent removing of (Required,Default) fields in HTTP PUT. Fixes #174

### LINKED ISSUE #174: PUT requests removes required fields with default values
If I PUT a new resource at `/resourcepath/myid`, It will set non-provided required fields with the specified `Default` values. If I do another PUT request against the same URL with the same body, required fields that are not part of the body gets cleared away...

I would instead have expected one of:
1. The request to be rejected with 422 due to missing required fields.
2. The request to fill in the missing fields with default values, so that calling a PUT request twice has the same effect as calling it once.

I believe 2 is the most correct / desirable thing to do, or to put it in other words; the least wrong thing to do. This would still "reset" fields that are not in the body to their default values, but that seams acceptable, and in some cases even desirable. It must of course be documented that this is what happens.

Case 1 is also much better than the current behaviour, but can not guarantee that a second PUT request will work just because the first one did, which if nothing else, is vert convenient.

Rest-layer versions:
- 326f8072a1edb7738b3e4bed8a573a0c2ffdfe29
- rs/rest-layer-mem@25a468b0b8d157b9cb8db4126e85b3311f80a537
- rs/rest-layer-mongo@544bb6a27b38cf3c14d36cef03367acfb94c894f

Relevant references:
- Section 9.6 PUT in https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.htm

Dragomir-Ivanov:


rest/method_item_put_test.go
```diff
@@ -294,6 +363,179 @@ func TestPutItem(t *testing.T) {
...
+			ExtraTest:    checkPayload("foo", "66", map[string]interface{}{"id": "66", "foo": "baz", "bar": "value"}),
+		},
+		`pathID:found,body:valid,default:missing`: {
+			Init: initForDefaultField(now, yesterday),
+			NewRequest: func() (*http.Request, error) {
+				body := bytes.NewReader([]byte(`{"foo": "baz"}`))
+				return http.NewRequest("PUT", "/foo/1", body)
+			},
+			ResponseCode: http.StatusOK,
+			ResponseBody: `{"id": "1", "foo": "baz"}`,
```
smyrman: I believe this request should set `"bar"` to `"default"` (no matter what the original value is).

It should probably be possibly be possible to clear via the new JSON PATCH request syntax, but for PUT, I think the result should be the same  for the first an second request, if possible.

rest/method_item_put_test.go
```diff
@@ -294,6 +363,179 @@ func TestPutItem(t *testing.T) {
...
+			NewRequest: func() (*http.Request, error) {
+				body := bytes.NewReader([]byte(`{"foo": "baz"}`))
+				return http.NewRequest("PUT", "/foo/1", body)
+			},
+			ResponseCode: http.StatusUnprocessableEntity,
+			ResponseBody: `{
+				"code": 422,
+				"message": "Document contains error(s)",
+				"issues": {
+					"bar": ["required"]
```
smyrman: Same here, I think the `"bar"` should be set to `"default"` here, so that if I repeat a PUT request twise, it's semantically the same (or very similar) to running it once.

However, if this proves hard to achieve, this functionality is a lot better than silently clearing the field and is preferable over the current way in which things works.

rest/method_item_put_test.go
```diff
@@ -294,6 +363,179 @@ func TestPutItem(t *testing.T) {
...
+			ExtraTest:    checkPayload("foo", "2", map[string]interface{}{"id": "2", "foo": "baz", "bar": "value"}),
+		},
+		`pathID:found,body:valid,required-default:delete`: {
+			Init: initForRequiredDefaultField(now, yesterday),
+			NewRequest: func() (*http.Request, error) {
+				body := bytes.NewReader([]byte(`{"foo": "baz"}`))
+				return http.NewRequest("PUT", "/foo/2", body)
+			},
+			ResponseCode: http.StatusOK,
+			ResponseBody: `{"id": "2", "foo": "baz", "bar": "default"}`,
```
smyrman: This looks like what I would expect for ID 1 as well. What's the difference between the two cases?

schema/schema.go
```diff
@@ -132,6 +132,8 @@ func (s Schema) Prepare(ctx context.Context, payload map[string]interface{}, ori
 				// previous value as the client would have no way to resubmit the stored value.
 				if def.Hidden && !def.ReadOnly {
 					changes[field] = oValue
+				} else if def.Required && def.Default != nil {
+					changes[field] = def.Default
```
smyrman: Should this be like this?

```go
} else if def.Default != nil {
    changes[field] = def.Default
}
```

schema/schema.go
```diff
@@ -227,7 +229,7 @@ func (s Schema) validate(changes map[string]interface{}, base map[string]interfa
 		}
 		// Check required fields.
 		if def.Required {
-			if value, found := changes[field]; !found || value == nil {
+			if value, found := changes[field]; !found || value == nil || value == Tombstone {
```
smyrman: 👍

rest/method_item_put_test.go
Dragomir-Ivanov: This is where the document is `found`. So `Default` field might have been deleted in a previous call, I don't think we need to reset it to Default. I believe (Default) should be applicable only for new documents.

rest/method_item_put_test.go
Dragomir-Ivanov: Actually this is kinda impossible to happen.  For (Required,Default) we can't have existing document without that field set ( ID:"1"). I did this test for completeness sake.

rest/method_item_put_test.go
Dragomir-Ivanov: In ID 1 (Requred,Default) field is not present (impossible to happen with `rest-layer` only), where in ID 2 it is present and eligible for deletion.

schema/schema.go
Dragomir-Ivanov: See first comment above.

rest/method_item_put_test.go
smyrman: I have the following concerns:
1. Just _clearing_ the field, makes it unsafe to repeat a PUT call (when we don't know if the original item has been created or not).
2. Keeping the _original_ value becomes very inconsistent with fields that don't have a default.

Consider the following use-case: A smart home application of some sort, starts up by reading a config file, and calling "PUT" to create several "lights" with values set in the config file. When a value is not set in the config file, we want to use the server defaults.

With the sematics as described in this test, the first time we start the application, it will work, but for the second boot, it will _clear_ the fields it set for the first start. Same for any "lights" or "devices" we add to the config file after that; they will only remain valid for the first boot, and then end up clearing default values.

PUT is meant to hold _replace_ semantics; we are not updating a resource at a given URL, we are simply replacing it. for selective update, there is PATCH. When doing so, setting default values makes sense in my mind. There then are some exceptions, such as read-only fields, e.g. "createdAt", that I think it makes sense to still persist.

Clearing some field with a default value, is an extreme special case, and even if you are allowed to do so via a future JSON PATCH request, a PUT request at the URL should still "replace" the original resource.

rest/method_item_put_test.go
```diff
@@ -294,6 +363,179 @@ func TestPutItem(t *testing.T) {
...
+			ExtraTest:    checkPayload("foo", "1", map[string]interface{}{"id": "1", "foo": "baz", "bar": "value"}),
+		},
+		`pathID:found,body:valid,default:delete`: {
+			Init: initForDefaultField(now, yesterday),
+			NewRequest: func() (*http.Request, error) {
+				body := bytes.NewReader([]byte(`{"foo": "baz"}`))
+				return http.NewRequest("PUT", "/foo/2", body)
+			},
+			ResponseCode: http.StatusOK,
+			ResponseBody: `{"id": "2", "foo": "baz"}`,
```
smyrman: I think here we should not clear  the `bar` value.

Consider this case:

A "home automation" application starts up by reading a list of "lights" from a config file, and it PUTs all the lights,so that it can later rely on them being there. We want the app to use the server defaults when there are no value configured in the config file. The first time we run the app, all the lights end up with the correct values, but whn we restart it, and rerun all the PUT requests, we will (unintentionally) end up clearing all non-required defaults.

PUT is meant to hold _replace_ semantics, so ideally, I think resetting the values to the _default_ is the correct thing to do, but since we won't/can't do that for read-only fields (e.g. createdTime), I can accept not doing it for _default_ fields neither. However, _clear_ fields with server-side defaults, no matter if they are required or not, may have odd side-effects in real applications as calling N PUT request has a semantically very different effect compared to calling just one.

Dragomir-Ivanov: Yes, I hear you. But then, what is the difference between (Default) and (Required,Default)? I guess (Default) can be explicitly deleted with PATCH(JSON-Patch), wheres (Required,Default) will get its Default value on delete?

rest/method_item_put_test.go
Dragomir-Ivanov: This is how `Default` is described in the README:
```
Default | 
The value to be set when resource is created and the client didn’t provide a value for the field. The content of this variable must still pass validation.
```
So it is tailored for resource creation, and its subsequent update is not mentioned. My code doesn't touch `ReadOnly` code path I presume. Deleting Read-Only field will bail out on Validation.

rest/method_item_put_test.go
Dragomir-Ivanov: I am puzzled how we should proceed with (Default) in this case. On one hand PUT is used to "PUT exactly this resource there!", then we set some fields under its feet, and for (Default) fields we can loose ability to delete them.

smyrman: > I guess (Default) can be explicitly deleted with PATCH(JSON-Patch), wheres (Required,Default) will get its Default value on delete?

I think that would make sense.

> On one hand PUT is used to "PUT exactly this resource there!", then we set some fields under its feet, and for (Default) fields we can loose ability to delete them.

Indeed:-) I think [this article](https://medium.freecodecamp.org/rest-is-the-new-soap-97ff6c09896d) says it in a good way:

> You want to use PUT to update your resource? OK, but some Holy Specifications state that the data input has to be equivalent to the representation received via a GET. So what do you do with the numerous read-only parameters returned by GET (creation time, last update time, server-generated token…)? You omit them and violate the PUT principles? You include them anyway, and expect an “HTTP 409 Conflict” if they don’t match server-side values (forcing you to then issue a GET...)? You give them random values and expect servers to ignore them (the joy of silent errors)? Pick your poison, REST clearly has no clue what a read-only attribute it, and this won’t be fixed anytime soon. Meanwhile, a GET is dangerously supposed to return the password (or credit card number) which was sent in a previous POST/PUT; good luck dealing with such write-only parameters too.

The thing is that no REST-API is able to completely follow the HTTP Standard and still be useful, and PUT is a particular tricky one. We actually used RPC calls in place of PUT in our own product, which allows us to design exactly the semantics we think makes sense for the app, and it allows us to do tailored things, like create multiple resources via one call...

But back to `PUT`, the truth is there is no 100% "correct" way to do it. Fields with default values and read-only fields just don't _fit_ with how PUT is  specified at all, but a REST API still need to set them in a way that's useful and at least follows some of the _principals_ of the HTTP standard while still retaining functionality.

Dragomir-Ivanov: Thinking a bit more @smyrman about your "lights" example, I believe you can freely use (Required,Default) and it will work the way you want it. Isn't it?

Dragomir-Ivanov: @rs Any input appreciated.

smyrman: >Thinking a bit more @smyrman about your "lights" example, I believe you can freely use (Required,Default) and it will work the way you want it. Isn't it?

Yes Setting all the fields with default to required would make it work.

>  EDIT: I am beginning to think rest-layer should not support (Default) alone. (Required) and (Required,Default) only.

Allowed or not, _PUT_ could/should _probably_ treat them equally...  Maybe it's a good idea to have a look at how JSON Schema is treating required fields with defaults. They have probably had a pretty similar discussion at some point...

@rs, yes your input is appreciated 🙂

smyrman: As an update, JSON Schema is extremely wage when [specifying](https://json-schema.org/latest/json-schema-validation.html#rfc.section.10.2) or [explaining](https://json-schema.org/understanding-json-schema/reference/generic.html#metadata) how default should work. I haven't found any JSON Schema discussions related to fields that are both required and has a default so far.

One thing to note though, is that JSON schema has `required` as a list of required properties in an `object`, and not as a property on each object field. The strength of having `requires` as a list, is that it allows setting `required` fields also for cases such as `schema.Dict`. I am considering this move in #77, but that just an FYI. For now, it is the way it is.

smyrman: I would really like to see it land, if you are able to. As for changes to do now, if you can add a description on how it works under different conditions to the PR, that would be the highest prority for me.

I think it would be nice to treat fields with `Default` equally to `Default` + `Required`, as the way things work would potentially not need to change for #77 later. Preferably I think this involves resting the field to Deefault, but if that makes e.g. JSON Patch not work as intended, then do what you think is best. The _most_ important part of this change, is that it no longer silently _clear_ required fields without erroring. For the rest of the PUT semantics, I actually don't mind to much, as long as the data corruption issue is resolved.

 > I am considering this move in #77, but that just an FYI. For now, it is the way it is.

This was just a note on the `Field.Required` property. Sorry if it was confusing.

Dragomir-Ivanov: > This was just a note on the `Required` being a property of `Field`. Sorry if it was confusing.
No, its fine. 

So are there any additional code changes you may want, or it is just the README? I would really love to see (Default) go, and leave only (Required) and (Required,Default).
@rs your input on this matter is greatly appreciated as well.

I will refactor the tests for PUT, and will move each field type into separate `Test*` function.

smyrman: I am happy with just the README changes.

Dragomir-Ivanov:
Done. Don't know if changed README as you imagined, but will fix :)
Merge conflicts are expected because my PRs touch adjustment blocks of text. Will fix.

Dragomir-Ivanov:
@smyrman Done.

Dragomir-Ivanov:
Conflict resolved.

README.md
```diff
@@ -1179,6 +1187,32 @@ Prefer: return=minimal
...
+The same as [GET](#get), except that it doesn't return any body.
+
+### GET
+Used to query a resource with its sub/embedded resources.
+
+### POST
+Used to create new resource document, where new `ID` is generated from the server.
+
+### PUT
+Used to create/update single resource document given its `ID`.\
```
smyrman: extra `\` and newline

README.md
```diff
@@ -1179,6 +1187,32 @@ Prefer: return=minimal
...
+Be aware when dealing with resource fields with `Default` set. Initial creation for such resources will set particular field to its default value if omitted, however on subsequent `PUT` calls this field will be deleted if omitted. If persistent `Default` field is needed use `{Required: true}` with it. 
+
+### PATCH
+Used to update/patch single resource document given its `ID`.
+
```
smyrman: Should we include the move the note on JSON-Patch here?

Dragomir-Ivanov: Yes, I guess we can. Will do it in a while.

Dragomir-Ivanov:
@smyrman Will do final squash when approved.

README.md
```diff
@@ -1179,6 +1186,44 @@ Prefer: return=minimal
...
+### PATCH
+Used to update/patch single resource document given its `ID`. REST Layer supports following update protocols:
+
+- Simple filed replacement [RFC-5789](http://tools.ietf.org/html/rfc5789) - this protocol will udpate only supplied top level fields, and will leave other fields in the document intact. This means that this protocol can't delete fields. Using this protocol is specified with `Content-Type: application/json` HTTP Request header.
+
+- [JSON-Patch/RFC-6902](https://tools.ietf.org/html/rfc6902) - When patching deeply nested documents, it is more convenient to use protocol designed especially for this. Using this protocol is specified with `Content-Type: application/json-patch+json` HTTP Request header.
+
+	If using `If-Match` concurrency control as described in the [data control and integrity section](#data-integrity-and-concurrency-control), you could potentially choose to calculate the body of new object client side. Note that the response body for a successful operation can be omitted by supplying a HTTP request header: `Prefer: return=minimal`.
```
smyrman: Is this indent by a tab for `If using...` accidental?

README.md
```diff
@@ -1179,6 +1186,44 @@ Prefer: return=minimal
...
+
+	```sh
```
smyrman: When using ```sh, no indentation is needed -- sorry if I missed this earlier.

README.md
Dragomir-Ivanov: It is tabbed because his paragraph goes to the last JSON patch section.

README.md
Dragomir-Ivanov: Same as above

README.md
smyrman: There is a preview function in the GitHub Code-Review (and in [VS Code](https://code.visualstudio.com/docs/languages/markdown)). It looks like it _works_, but I think it' looks a bit weird to have two paragraphs with newline spacing within a bullet-point. Would probably look better to just use paragraphs, or to some extent even tables.

I noticed just now, that there is another section called "Modes" that lists the HTTP Methods already. Should probably go over the README at some point and consolidate some sections...

For now, could you perhaps make the If-Match paragraph on independent of the bullet-point, and indent it to the left? I feel the text as it stands applies to both cases anyway; replacing a single-top level field is even easier to calculate client-side than to apply a JSON-Patch.

Same for the code block.

Dragomir-Ivanov: Yes, you are actually right, it relates to both ways of patching. Done.
VSCode has build-in Markdown preview, and I am using it to verify my changes. I think it matches GHs Markdown processor entirely.

PR #207, opened 2018-11-28
Add JSON-Patch support. Closes #145

### LINKED ISSUE #145: REST PATCH json protocol
Currently `rest-layer` implements HTTP PATCH method in a very crude way, i.e. it only replaces top level fields. How about adding something more advanced like: http://jsonpatch.com/ 
 https://tools.ietf.org/html/rfc6902

It can deeply patch JSON documents, and is supported by many languages, Go included.

Dragomir-Ivanov:


README.md
```diff
@@ -1235,6 +1237,20 @@ When a validator implements this interface, the method is called with the field'
 
 See [schema.IP](https://godoc.org/github.com/rs/rest-layer/schema#IP) validator for an implementation example.
 
+## JSON-Patch
+
+When patching deeply nested documents, it is more convenient to use protocol designed especially for this, instead
+of top level document field replacement as described in [RFC-5789](http://tools.ietf.org/html/rfc5789). For this purpose `rest-layer` implements [JSON-Patch/RFC-6902](https://tools.ietf.org/html/rfc6902). In addition JSON-Patch allows field deletion, which is not possible with the simple field replacement.
+
+Consider using JSON-Patch always with `If-Match`, to be sure which document is actually patched. In that case the updated document is already known, so HTTP responce body may be omitted via: `Prefer: return=minimal` HTTP requres header.
```
smyrman: minor typo: requres => request

The `If-Match` concurrency control is already mentioned elsewhere https://github.com/rs/rest-layer#data-integrity-and-concurrency-control. We should consider how much of that we want to repeat. A suggestion:

> If using `If-Match` concurrency control as described in the [data control and integrity section]( https://github.com/rs/rest-layer#data-integrity-and-concurrency-control), you could potentially choose to calculate the body of new object client side. Note that the response body for a successful operation can be omitted by supplying a HTTP request header: `Prefer: return=minimal`.

That said, I suppose this header can also be used for other requests? At least POST, PUT and PATCH without JSON-Patch. Is there some common place we could mention this in the docs?

Other than that, field selection works like for any other request (GET, POST, PUT), and could be used for limiting the response size. I don't know if it's worth mentioning though.

rest/method_item_patch.go
```diff
@@ -64,6 +125,11 @@ func itemPatch(ctx context.Context, r *http.Request, route *RouteMatch) (status
 		e = NewError(err)
 		return e.Code, nil, e
 	}
+
+	if isNoContent(r) {
+		return 204, nil, nil
+	}
+
```
smyrman: To include an E-tag in the response, an item must be returned:
https://github.com/rs/rest-layer/blob/48c11e50091e8c988780853a3aaa8af25cf98d42/rest/response.go#L74

Perhaps you can instead replace:
https://github.com/rs/rest-layer/blob/48c11e50091e8c988780853a3aaa8af25cf98d42/rest/handler.go#L51

with:
```go
skipBody := isNoContent(r)
```

Require you to handle the `r.Method == "HEAD"` as well in the `isNoContent` implementation. This header checks would now work for all update requests. Should also probably ignore the headers for `r.Method == "GET"`.

rest/method_item_patch.go
```diff
@@ -2,20 +2,59 @@ package rest
...
 )
 
+func isJSONPatch(r *http.Request) bool {
+	if ct := r.Header.Get("Content-Type"); ct != "" && strings.TrimSpace(strings.SplitN(ct, ";", 2)[0]) == "application/json-patch+json" {
+		return true
+	}
+	return false
+}
+
+func isNoContent(r *http.Request) bool {
```
smyrman: Suggesting some changes, see comment for where it's called from in the code.

smyrman (changes requested):
Really solid work, but some small comments for improvement. @rs -- are you happy we add the `jsonpatch` dependency?

rest/method_item_patch.go
Dragomir-Ivanov: Oh yes, totally forgot about ETags. I will add at least one check for PATCH to check about the ETag in the response. Also I guess I will need to add some tests for other HTTP methods regarding `skipBody`. Thank @smyrman good catch.

README.md
Dragomir-Ivanov: I wanted to highlight what is the intended use for patching with JSON-Patch, alongside `If-Match`. I don't recall any common place to describe common HTTP methods functionality. I will see where I can put this.

rest/method_item_patch.go
smyrman: `HEAD` requests are allowed to contain a body:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD

> A response to a HEAD method should not have a body. If so, it must be ignored. Even so, entity headers describing the content of the body, like Content-Length may be included in the response. They don't relate to the body of the HEAD response, which should be empty, but to the body of similar request using the GET method would have returned as a response.

However, failed PATCH,PUT or PAST requests should include an error. Maybe we need two parameters?

```go
skipError := r.Method == "HEAD"
skipOKBody := isNoContent(r)
```

Dragomir-Ivanov: But what if `HEAD` has an error. Shouldn't we get the formatted error with appropriate `message` generated by `rest-layer`.

Dragomir-Ivanov:
Made needed changes. Split the `JSON patch` and `Prefer` in separate commits.

rest/method_item_patch.go
Dragomir-Ivanov: Done

smyrman (commented):
> so I wonder if we need ever change the RC to 204. Any advice @smyrman @rs ?

Don't really have any advice.

It feels explicit to give a 204 when the body has been omitted, so I think the way your PR currently handles (and documents) it is good. That you don't replace 201 with 204, but only 200, I think is also probably reasonable.

README.md
```diff
@@ -52,9 +52,12 @@ The REST Layer framework is composed of several sub-packages:
 - [Authentication & Authorization](#authentication-and-authorization)
 - [Conditional Requests](#conditional-requests)
 - [Data Integrity & Concurrency Control](#data-integrity-and-concurrency-control)
+- [HTTP Headers Support](#http-headers-support)
+	- [Prefer](#prefer)
```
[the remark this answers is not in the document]
Dragomir-Ivanov: Oh yes, sorry. I guess VSCode detects tab?
How about adding `.editorconfig` file, where we can set identification gradually down to a specific file.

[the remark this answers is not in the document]
Dragomir-Ivanov: Fixed and rebased. For leaving 201, I wanted to client to know that resource still was created.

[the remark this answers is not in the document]
smyrman: `.editorconfig` files are awesome -- but I will let @rs comment at it as this file is almost just `.go`. As for the README, it currently uses space for _most_ indentation, but tabs for copy-pasted in Go code indentation; that's presumably a bit hard to configure in any editor.

smyrman (commented):
I will use some more time later to read through the tests before I aprove, but this seams to be really good!

smyrman (approved):
LGTM. @rs, are you OK with depending on github.com/evanphx/json-patch?

rs:
sounds good

PR #209, opened 2018-11-30
Adding go.mod support. Closes #193

### LINKED ISSUE #193: Add go.mod file (go module support)
Experimental Go  module support has landed in the Go 1.11 beta https://groups.google.com/forum/#!msg/golang-dev/a5PqQuBljF4/61QK4JdtBgAJ. This introduces 

It would be nice to add a `go.mod` file with dependency constraints for rest-layer around the time when Go 1.11 is released, and of course to have it in place before the v0.2 release.

This should be a relatively easy task in itself, although there may be some minor issues here and there.

Dragomir-Ivanov:


Dragomir-Ivanov:
Hmm, I have never worked with `mod` before, so I read some for this patch. 
I find this explanation:
```
The module download is split into two parts: downloading the go.mod and downloading the actual code. If you have dependencies only needed for tests, then they will show up in your go.mod, and go get will download their go.mods, but it will not download their code. The test-only dependencies get downloaded only when you need it, such as the first time you run go test.
```
source: https://github.com/golang/go/issues/26913#issuecomment-411976222

smyrman:
We could potentially exclude _example_ dependencies by adding a `go.mod` file to the `examples` directory and running `go build` and/or `go mod tidy` on the relevant packages.

smyrman:
Maybe I can just do it :-)

smyrman:
Moved to #210.

PR #210, opened 2018-11-30
Converting to go 1.11 modules

smyrman:
Removes vendored version of hysterix.

Initattes two new go module roots:
- One for the git root
- One for the examples/ sub-directory

Resolves #209


examples/go.mod
```diff
@@ -0,0 +1,15 @@
+module github.com/rs/rest-layer/examples
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/justinas/alice v0.0.0-20171023064455-03f45bd4b7da
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/rs/cors v1.6.0 // indirect
+	github.com/rs/rest-layer v0.1.0
```
smyrman: Will update this to latest master.

smyrman: Used a "relace" directive instead.

smyrman:
The example_test.go was moved to the `examples/` folder to get the mod file to detect less dependencies. The `examples/` folder got it's own mod file, with a replace directive for the rest-layer dependency.

Any oppinions on this approach @rs?

rs:
I don't think you need to do that. Just do no run go test before committing the go.mod and test deps won't be added.

Dragomir-Ivanov:
```
go mod init github.com/rs/rest-layer; go mod tidy
```
Always includes test deps, without any `go test` run, after the above command.

smyrman:
As for adding a separate go module root in the `examples/` directory, that makes (as far as I know) anything depending on `rest-layer`, or commands run against `rest-layer`, including `go test ./...`, completely ignore  the `examples/` directory. I think this is reasonable - the dependencies found in the example are completely irrelevant for rest-layer as a library - but I can't claim there is a precedence in the community for doing it in this way yet.

smyrman:
> As for moving the `example_test.go` to `examples/`, a _better_ alternative for keeping the amount of dependencies down in the main module, would be to rewrite it to only depend on the standard library (e.g. http.Handler, the default logger etc).

I have done an attempt on this. Kept the `cors` dependency as it's hard to avoid, and is already included in another test.

smyrman:
I will go ahead and merge this.

PR #212, opened 2018-12-02
Make `Prepare` method of `schema.Or/schema.And` non-pointer receiver.…

Dragomir-Ivanov:
… Artifact from 4e99227839137d613766f0afa0f4ada81ba2728e

### DIFF
--- a/schema/query/predicate.go
+++ b/schema/query/predicate.go
@@ -83,8 +83,8 @@ func (e And) Match(payload map[string]interface{}) bool {
 }
 
 // Prepare implements Expression interface.
-func (e *And) Prepare(validator schema.Validator) error {
-	return prepareExpressions(*e, validator)
+func (e And) Prepare(validator schema.Validator) error {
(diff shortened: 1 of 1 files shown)


smyrman:
Thanks a lot:-)

Let me do some more testing/thinking about this before we merge.

As I mentioned on Slack, the good thing about letting this _not_ be pointer receiver, is that you can `append` and `range` over it without dereferencing.

There is however a case for keeping it as is as well, and that is you are now granted to only get passed _one_ possible type of `*query.And` and `*query.Or`, but if moving the Prepare statement to non-pointer reciever, we must have code that actually accepts _both_ in e.g. switch case environments.

smyrman:
I am inclined to close this; switch cases seam a common pattern when handling with query.Expressions, and having two types that are not enforced as pointers, would just make them appear s both.

Thanks though.

Dragomir-Ivanov:
Fair enough.

PR #215, opened 2018-12-09
Add field projection to sub-resource/referenced resource queries.

Dragomir-Ivanov:
These potentially can be used in `resource hooks` and `Storer` interface implemeters.
No tests breakage. No new test cases added, since I couldn't find this functionality tested in current tests. May add some tests about these in the future.

### DIFF
--- a/schema/query/projection_evaluator.go
+++ b/schema/query/projection_evaluator.go
@@ -79,7 +79,10 @@ func evalProjection(ctx context.Context, p Projection, payload map[string]interf
 					}
 				} else if ref, ok := def.Validator.(*schema.Reference); ok {
 					// Execute sub-request in batch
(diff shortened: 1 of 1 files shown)


Dragomir-Ivanov:
Okay, I may remove that part of the commit message.

Dragomir-Ivanov:
@smyrman Can we merge that, because it potentially will interfere with #222 .

rs:
I think the doc is still relevant for storers. It can make sense for hooks to evaluate projections though.

Dragomir-Ivanov:
Guys, pressed `Update branch` button here in GitHub, and it generated one "Merged" git commit. I can remove it if necessary.

smyrman:
Yes, let's merge it -- but I need you to do a final rebase before I am allowed to merge it, can you do that?

PR #216, opened 2018-12-11
schema: AnyOf FieldComparator implementation + #185

### LINKED ISSUE #185: schema: Let AnyOf/AllOf implement the FieldGetter interface
# Background
The FieldGetter interface basically allows the `schema/query` package to lookup fields specified in a `filter`, and see if the fields exist and if they are marked as searchable. It also allows validating the qurery _value_ although that feature is currently partially broken (see #170). `schema.Schema`, `schema.Array`,  `schema.Object` and `schema.Dict` implements this interface today.

# The issue
The purpose of this issue is straight forward; if there is one or more implementation of a `FieldGetter` interface added as members in a `schema.AnyOff` or `schema.AllOff` `FieldValidator`, allow the fields to be found.

# Solution
The implementation for this is probably straight farward, and should be the same for both `schema.AnyOf` and `schema.AllOff`: Get and return the first matching field:

```go
// GetField returns the first field that match name or nil.
func (v AnyOff) GetField(name string) *Field {
	for _, vv := range v {
		fg, ok := vv.(FieldGetter)
		if !ok {
			continue
		}
		if f := fg.GetField(name); f != nil {
			return f
		}
	}
	return nil
}
```

smyrman:
This PR groups a few commits that does the following things in the schema package:
- Move the tests of all the `FieldComparator` implementations to test only the public interface.
- Extends comparability to include `schema.AnyOf`.
-  Improve `schema.AnyOf` errors.
-  Resolve #185 (let `AnyOf` / `AllOf` implement the `FieldGetter` interface).

## Copy of commit messages (latest first):

commit 1d4ba96a7937d8d065525f6181365e403caa1c02 (origin/smyrman/anyof-cmp)
Author: Sindre Røkenes Myren <sindre@searis.no>
Date:   Tue Dec 11 22:08:15 2018 +0100

    travis: fix go modules bug
    
    Travis is running pre-scripts to go get everything without go modules
    enabled. Moving environment variable to `env` section.
    
    There was a typo in the scripts section when setting GO111MODULE.

commit 7ed6b21b824c2321d8ebd995cb870e6b90395743
Author: Sindre Røkenes Myren <sindre@searis.no>
Date:   Tue Dec 11 01:40:30 2018 +0100

    schema: Allow AnyOf to implement FieldComparator
    
    This changes the interface of FieldComparator so that it's possible for
    FieldValidator types to determine, either run-time or compile-time, if
    the validator is comparable.
    
    AnyOf has been modified to Allow comparisons if one of it's
    sub-validators allows it.
    
    Changes the comparable types Float, Integer, AnyOf and Time to test only
    the public interface; testing the private interface gives little value,
    as it should be allowed to change without test updates.

commit 749c6a3a4256717f5b6f42454faba60dd4158996
Author: Sindre Røkenes Myren <sindre@searis.no>
Date:   Tue Dec 11 01:30:47 2018 +0100

    schema: Let AnyOf/AllOf implement FieldGetter
    
    Lets AnyOf and AllOf implement the FieldGetter interface.

commit 52c1b21ca7291afe5c6ca9e58c854d9359c3b69a
Author: Sindre Røkenes Myren <sindre@searis.no>
Date:   Tue Dec 11 01:29:38 2018 +0100

    schema: better errors for AnyOf/AllOf + cleanups
    
    AnyOf and AllOf have received better error messages and doc-strings.
    They have also been changed so that they will accept any value if there
    are no members.
    
    Clean up mergeFieldErrors to be easer to read; Go allows appending to
    nil slice and expansion of slices in append, so there is no need for
    the if branch and nested for-loop that was there before.
    
    Simplify error handling in Object.Validate slightly. Moves the ErrorMap
    type from object.go to errors.go and add some helper methods.


schema/anyof.go
```diff
@@ -68,3 +73,27 @@ func (v AnyOf) Serialize(value interface{}) (interface{}, error) {
 
 	return value, nil
 }
+
+// LessFunc implements the FieldComparator interface, and returns either the
+// first non-nil LessFunc or nil.
+func (v AnyOf) LessFunc() (LessFunc, bool) {
```
smyrman: PS! There is a noticeable limitation with this design. Consider the case:

```go
"myField": {
  Default: null,
  Validator: &schema.AnyOf{
    &schema.String{},
    &schema.Float{},
  },
}
```

The query parse step would now allow a query `{myField:{&gt:null}}`, even though `null` is not strictly allowed for a compare operation; that is because `AnyOf` actually should use different criteria (consider only `schema.FieldCompirator` implementations), when validating a query value for a conspirator operator.

PErhaps we can allow this limitation for now (and for v0.2). If I have time during Christmas, I am planning to have a look at forking the `schema` package to a new repo and try to come up with a design that can address #77. I want to do this in a separate repo, so I can experiment with designs without worrying about breaking things.

Dragomir-Ivanov: Thanks, it seems acceptable.

smyrman:
@rs / @Dragomir-Ivanov, this is now ready for review.

Dragomir-Ivanov:
Except couple of code comments, LGTM.

schema/anyof_test.go
```diff
@@ -126,7 +132,7 @@ func TestAnyOfQueryValidate(t *testing.T) {
 			Name:      `{Bool,Bool}.Validate("")`,
 			Validator: schema.AnyOf{&schema.Bool{}, &schema.Bool{}},
```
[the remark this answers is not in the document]
smyrman: First of all, no it doesn't make sense in real life, secondly, I don't think it should error during compilation; that would just lead to unpredictable semantics, not to mention it would be hard to implement. Repeating the same `FieldValidator` _type_ with different parameters certainly is valid. 

If we compare it to JSON Schema, it does not really make sense to repeat the same schema in an anyOf clause, but it's still 100% valid according to the spec.

For the sake of this test, I think it's fine as it is.

smyrman:
Rebased just now. @rs - do you have some time to review this?

schema/error.go
```diff
@@ -0,0 +1,79 @@
...
+	return e
+}
+
+// ErrorSlice contains a concatenation of several errors.
+type ErrorSlice []error
+
+// Add adds an error to e and returns a new slice. If err is ErrorSlice, it is
+// extended so that all the elements in err are apppended to e. If err is nil,
+// then no error is appended. Note that calling Add on a nil slice is valid.
+func (e ErrorSlice) Add(err error) ErrorSlice {
```
rs: As it works the same way as `append`, why not calling this method `Append` instead of `Add`?

schema/error.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	sort.Strings(errs)
+	for i, key := range errs {
+		errs[i] = fmt.Sprintf("%s is %s", key, e[key])
+	}
+	return strings.Join(errs, ", ")
+}
+
+// Extend copies all errors from err to e.
+func (e ErrorMap) Extend(err ErrorMap) {
```
rs: Shouldn't this be called `Merge`?

schema/error.go
```diff
@@ -0,0 +1,79 @@
...
+	for k, v := range err {
+		e[k] = append(e[k], v...)
+	}
+}
+
+// Tidy returns an un-typed nil if e is empty, and an ErrorMap otherwise.
+func (e ErrorMap) Tidy() error {
```
rs: That's a strange API. Why do you need that? Couldn't we just expose a `Len` method and let the caller do what's appropriate?

schema/object.go
```diff
@@ -30,10 +27,10 @@ func (v Object) Validate(value interface{}) (interface{}, error) {
 		return nil, errors.New("not an object")
 	}
 	dest, errs := v.Schema.Validate(nil, obj)
-	if len(errs) > 0 {
-		var errMap ErrorMap
-		errMap = errs
-		return nil, errMap
+	if err := ErrorMap(errs).Tidy(); err != nil {
```
rs: Per my comment above, I would find it easier to read if it was:

```go
if len(errs) > 0 {
    return nil, ErrorMap(errs)
}
```

schema/error.go
smyrman: I agree, it's strange. Will change it. Len could work.

smyrman: Fixed

schema/error.go
smyrman: Fixed

schema/error.go
smyrman: Fixed

schema/object.go
smyrman: Fixed

smyrman:
@rs, apologies for my very slow response on this one. I have now updated the PR with your suggested changes.

Dragomir-Ivanov:
Nice work @smyrman . Everything looks good, however I find it strange when we get the first `lessFunc` for `schema.AnyOf`. Is this right? It seems that `shema.AnyOf` is mainly used for `shema.AnyOf{shema.Null, shema.XXX}`. Can't we just add a field type `nullable`, and fix this?

smyrman:
We have many use-case for `schema.AnyOff` besides `shema.AnyOf{shema.Null, shema.XXX}`, but the place where we need the less function to be implemented is indeed such cases.

Here are some examples of what we have (including some custom FieldValidators):

```go
Validator: &schema.AnyOf{
	&schema.Null{},
	&rsc.Duration{Min: 2 * time.Millisecond, Max: 365 * 24 * time.Hour},
}
```
```go
Validator: &schema.AnyOf{
	&schema.Null{},
	rsc.UTCTruncTime{Truncate: time.Hour * 24},
}
```

But even if we removed the `Null` type and added a `Nullable` filed (for convenience or as a short-hand), it would still be completely valid to define a schema like this:

```go
Validator: &schema.AnyOf{
	&schema.Interger{Boundaries:&schema.Boundaries{Min:-5,Max:0}},
	&schema.Interger{Boundaries:&schema.Boundaries{Min:5,Max:10}},
}
```

Weird perhaps, but valid.

I think it would be good to aim for feature completeness in what's supported for AnyOf, AllOf so that they are more useful and more usable.

I am happy to discuss if adding the Less function has any unwanted side-effects that would be best to avoid.

smyrman:
rebased on master (no new changes)

smyrman:
thx 🙇

PR #217, opened 2018-12-12
Update Storage Handlers list

apuigsech:
Add generic SQL Storage Handler and remove SQLite3.


smyrman:
@rs, will just merge this.

PR #221, opened 2018-12-22
Added Kubernetes ConfigMap storage backend to README

### LINKED ISSUE #211: Kubernetes ConfigMap backend
Hi there,

I'd just like to let you guys know that I build a backend for REST Layer that stores documents in Kubernetes configmaps: https://github.com/Segence/rest-layer-kubernetes-configmap

robvadai:
Resolves #211

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,7 @@ As REST Layer is a simple `net/http` handler. You can use standard middleware to
 - [x] [ElasticSearch](http://github.com/rs/rest-layer-es)
 - [x] [SQL](https://github.com/apuigsech/rest-layer-sql) (third party)
 - [x] [Google Datastore](https://github.com/ajcrowe/rest-layer-datastore) (third party)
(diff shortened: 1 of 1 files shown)


apuigsech:
Does it make sense that Kubernetes ConfigMap is a generic REST Layer storage handler? Maybe I am wrong, but It's not so clear to me and I am going to try to argue.

I think that a Storage Handled should allow to store any kind of schema (of course, considering exceptions and characteristics of each storage backend), and in the case of ConfigMap the schema is defined by ConfigMap itself. In the case of this handled is defined on "configmap.ConfigMapSchema".

I think this is a great example about how to connect REST Layer to a backend that has it's own Schema, but I don't see it as a generic Storage Handler in the same category than the rest of elements on the list. Maybe a section different section "use cases" or similar can be created to link that kind of thinks.

Opinions?

smyrman:
Probably the third-party storage back-ends that has limitations can get their own section? E.g the SQL back-end also have limitations in which fields you can query from what I can see, I.e. not sub-schemas.

PR #224, opened 2019-01-02
Issue 222 223

Dragomir-Ivanov:


schema/query/projection_evaluator_test.go
```diff
@@ -186,6 +186,48 @@ func TestProjectionEval(t *testing.T) {
...
+			`{"reference":"100"}`,
+			nil,
+			`{"reference":null}`,
+		},
+		{
+			"ReferenceArray-non-existant",
+			`reference{name}`,
+			`{"reference":["100"]}`,
+			nil,
+			`{"reference":[]}`,
```
smyrman: It feels a bit more consistent to me to expand this as `"refrence": [null]`; then the behavior of an array of references is more consistent with the single reference field case. I believe keeping `null` should already be  behavior of `MultiGet`?

smyrman (commented):
Sorry for the delay.

schema/query/projection_evaluator_test.go
smyrman: Simmilar,

`{"reference":["2","100","3"]}`,

--->

`{"reference":[{"name":"second"},null,{"name":"third"}]}`

Ideally, there would be a way to filter away `null` as well, but I feel weather or not a UI wold want to highlight dead references is not up to the API to decide?

smyrman (approved):

schema/query/projection_evaluator_test.go
smyrman: I actually don't feel too strongly about it though... Perhaps "including `null` is the least common case that needs to be made explicit if we are to support it.

Approved as is.

smyrman:
@Dragomir-Ivanov, is there any more to this, or can I just merge it?

schema/query/projection_evaluator_test.go
Dragomir-Ivanov: Thanks for the review Sindre, I didn't thought about this. I can include `null`s in place of absent resources, and actually this would be good idea. Say one requests a sub-resource, but make a projection, and doesn't include `id` in the said projection. Then he does have no way of knowing which resource item is missing. 
```
?fields=reference{name,age}
```
Without `id` in the resource item, the client will not be able to decide which resource got fetched and which not. I understand that this is a corner case, but still good to have.

Dragomir-Ivanov:
@smyrman Did one comment on the review. Please advice.

schema/query/projection_evaluator_test.go
Dragomir-Ivanov: However, looking at the current `rest-layer` implementation, it would be hard to accomplish this. `MultiGet` will not be triggered in the:
```
if isArray {
	e = &In{Field: "id", Values: v}
```
becase this is what guards it:
```
if len(q.Predicate) == 1 {
		if eq, ok := q.Predicate[0].(*Equal); ok && eq.Field == "id" {
			// Make an optimization for query on a single id so we can coalese them into a single request.
			id := eq.Value
			for _, r := range rbr.requests {
				if r, ok := r.(*referenceMultiGetRequest); ok && r.resourcePath == resourcePath {
					r.add(id, handler)
					return
				}
```
Perhaps, we can chat a little on Slack, for me to understand what exactly `MultiGetter` actually does.

Dragomir-Ivanov: Tried to patch `referenceBatchResolver` to use `MultiGet` with `query.In` but no success. It seems that I don't understand `Multigetter` enough. I guess we can merge as it is, so at least users can request `id` field of the referenced resource, and do item discovery themselves.

Dragomir-Ivanov: Thinking a bit more, I conclude that client shouldn't care if referenced item is present or not, so no `null` needed in place of missing items. Closing the review, please merge as is.

PR #228, opened 2019-01-09
Implements '*'(expand all) field selector. Closes: #147

Dragomir-Ivanov:


schema/query/projection_evaluator.go
```diff
@@ -39,25 +39,63 @@ func (p Projection) Eval(ctx context.Context, payload map[string]interface{}, rs
...
+				proj = append(proj, ProjectionField{Name: fn, Children: starChildren})
+			}
 		}
 	}
+	return proj, nil
+}
+
+func evalProjection(ctx context.Context, p Projection, payload map[string]interface{}, fg schema.FieldGetter, rbr *referenceBatchResolver) (map[string]interface{}, error) {
+	res := map[string]interface{}{}
+	resMu := sync.Mutex{} // XXX use sync.Map once Go 1.9 is out
```
smyrman: Cam remove FIXME. SyncMap also probably isn't fit for purpose here.

Dragomir-Ivanov: Removed in separate commit.

Dragomir-Ivanov: @smyrman Did you noticed that I have changed method parameter from `validator schema.Validator` to `fg schema.FieldGetter`?

Dragomir-Ivanov:
@smyrman This is whole revamped version if the functionality. There is a lot to digest, so take your time with the review.

Dragomir-Ivanov:
It may be worth checking out my repo, and review with:
```
git diff b810f68..HEAD
```
Because this patch, reverts part of my previous work on arrays of `schema.Reference` that is already upstream.

schema/query/projection_evaluator.go
```diff
@@ -79,47 +214,57 @@ func evalProjection(ctx context.Context, p Projection, payload map[string]interf
 					}
 				} else if ref, ok := def.Validator.(*schema.Reference); ok {
```
smyrman: nicer code with a type-switch btw.

```go
switch ref := def.Validator.(type) {
case *schema.Reference:
     // ref will be of type  *schema.Reference in this block
case *schema.Array:
    //  ref will be of type  *schema.Array in this block
case FieldGetter:
    //  ref will be of type  FieldGetter in this block.
default:
   return nil, fmt.Errorf("%s: field has no children", pf.Name)
}
```

Slightly better to have `FieldGetter` than `*schema.Dict`; this way we can also handle `*schema.Object`. `*schema.Array` is a bit different, unfortnunatly, and should probalby implement som other interface, e.g.

```go
type IndexGetter interface{
    GetField(i int) (Field, bool)
}
```

But we can leave that for later. A static `*schema.Array` check is fine for now.

I must also note though that this function in general (as it was written before this review), is too long and to deeply nested to my taste, and could benefit from some helper functions, or re-evaluation of the if/else logic/nesting. That's for later though.

schema/query/projection_evaluator.go
```diff
@@ -39,25 +39,160 @@ func (p Projection) Eval(ctx context.Context, payload map[string]interface{}, rs
...
+					return fmt.Errorf("%s: error applying projection on sub-field item #%d: %v", pf.Name, i, err)
+				}
+			}
+			if v, err = resolveFieldHandler(ctx, pf, def, payloads); err != nil {
+				return fmt.Errorf("%s: error resolving field handler on sub-field: %v", name, err)
+			}
+			vv := v.([]map[string]interface{})
+			resMu.Lock()
+			for _, item := range vv {
+				res = append(res, item)
```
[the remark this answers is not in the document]
Dragomir-Ivanov: Can't do. `res` is `[]interface{}`, `vv` is `[]map[string]interface{}`.

schema/query/projection_evaluator.go
Dragomir-Ivanov: Made suggested type switch in `evalProjectionArray`.

schema/query/projection_evaluator.go
```diff
@@ -135,12 +136,12 @@ func evalProjectionArray(ctx context.Context, pf ProjectionField, payload []inte
 			}
 			res = append(res, v)
 		}
-	} else if array, ok := validator.(*schema.Array); ok {
+	case *schema.Array:
```
smyrman: Since `*schema.Array` implements `FieldGetter`, the `case *schema.Array` need to be placed first or we would always end up at the FieldGetter case and never hit the `*schema.Array` case.

smyrman: https://play.golang.org/p/nYdtczbVxYJ

schema/query/projection_evaluator.go
```diff
@@ -79,47 +197,57 @@ func evalProjection(ctx context.Context, p Projection, payload map[string]interf
...
+						// Return `null` for missing result instead of empty object
+						if m, ok := v.(map[string]interface{}); ok && len(m) == 0 {
+							v = nil
+						}
 						resMu.Lock()
 						res[name] = v
 						resMu.Unlock()
 						return nil
 					})
+				} else if dict, ok := def.Validator.(*schema.Dict); ok {
```
smyrman: Here we could perhaps also use `FieldGetter` instead of `*schema.Dict`. Again, the `*schema.Array` check must happen before the `FieldGetter` check then, as `*schema.Array` implements `FieldGetter`.

schema/query/projection_evaluator.go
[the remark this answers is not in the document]
smyrman: Ah, ok.

schema/query/projection_evaluator.go
smyrman: This is legacy reasons, explained near the top of #77

Dragomir-Ivanov: Okay, so the latest patch will do for now.

PR #233, opened 2019-01-29
Fix creating sub-resource items deeper than 2nd level

Dragomir-Ivanov:
Fixes #231

### DIFF
--- a/rest/resource_path.go
+++ b/rest/resource_path.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 
 	"github.com/rs/rest-layer/resource"
+	"github.com/rs/rest-layer/schema"
 	"github.com/rs/rest-layer/schema/query"
 )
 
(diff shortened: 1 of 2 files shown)


rest/resource_path.go
```diff
@@ -152,10 +153,28 @@ func (p ResourcePath) Path() string {
...
+	targetResource := p[len(p)-1].Resource
+	targetFields := targetResource.Schema().Fields
+
 	v := map[string]interface{}{}
 	for _, rp := range p {
+		include := false
 		if _, found := v[rp.Field]; !found && rp.Value != nil {
-			v[rp.Field] = rp.Value
+			if def, ok := targetFields[rp.Field]; ok {
+				if ref, ok := def.Validator.(*schema.Reference); ok && ref.Path == path {
```
smyrman: This hard-ceded check of the type is a bit unfortunate, but I suppose it's ok, at the very least for now.

PR #235, opened 2019-02-08
Fix PUT,PATCH on read-only fields

### LINKED ISSUE #232: PUT,POST,PATCH with original values gives an error
Working with PATCH, I hit an issue doing updates onto resources:
This is what I do:
```
GET http://localhost:8080/api/visits/5a4d69264e2127a7118b4c51

Result:
{
  "appointmentDT": "2018-01-05T18:00:00+02:00",
  "clientId": "5a4d63fe41095f105dcd49b2",
  "createdDT": "2018-01-04T01:37:18.079+02:00",
  "duration": 180,
  "id": "5a4d69264e2127a7118b4c51",
  "name": "Dryer, Refrigerator, Microwave Diagnose",
  "serviceId": "5a4d63fe41095f105dcd49b6",
  "technitian": "Asparuh Vasilev",
  "zipCode": "89107"
}
```
Then I try to update with the same exact data:
```

PUT http://localhost:8080/api/visits/5a4d69264e2127a7118b4c51

{
  "appointmentDT": "2018-01-05T18:00:00+02:00",
  "clientId": "5a4d63fe41095f105dcd49b2",
  "createdDT": "2018-01-04T01:37:18.079+02:00",
  "duration": 180,
  "id": "5a4d69264e2127a7118b4c51",
  "name": "Dryer, Refrigerator, Microwave Diagnose",
  "serviceId": "5a4d63fe41095f105dcd49b6",
  "technitian": "Asparuh Vasilev",
  "zipCode": "89107"
}

Result:
{
  "code": 422,
  "issues": {
    "createdDT": [
      "read-only"
    ],
    "id": [
      "read-only"
    ]
  },
  "message": "Document contains error(s)"
}
```
This shouldn't be the case since, I am putting exactly the same data, and removing `read-only` fields from the WEB UI, seems tedious and overkill. I made a simple patch to trace the issue:
```
func (s Schema) Prepare(......

			if found && (!oFound || !reflect.DeepEqual(v, oValue)) {
				fmt.Printf("field=%s type=%T value=%#v  otype=%T ovalue=%#v \n", field, value, value, oValue, oValue)
				changes[field] = value
			}
```
Which gives following trace:
```
field=appointmentDT type=string value="2018-01-05T18:00:00+02:00"  otype=time.Time ovalue=time.Time{wall:0x0, ext:63650764800, loc:(*time.Location)(0xb47380)} 
field=serviceId type=string value="5a4d63fe41095f105dcd49b6"  otype=bson.ObjectId ovalue="ZMc\xfeA\t_\x10]\xcdI\xb6" 
field=id type=string value="5a4d69264e2127a7118b4c51"  otype=bson.ObjectId ovalue="ZMi&N!'\xa7\x11\x8bLQ" 
field=createdDT type=string value="2018-01-04T01:37:18.079+02:00"  otype=time.Time ovalue=time.Time{wall:0x4b571c0, ext:63650619438, loc:(*time.Location)(0xb47380)} 

```
This indicates that for some types, original values are in DB format, but incoming are in text format. Later in `func (s Schema) Validate(......` conversion is made, but this is too late.

Can I propose following change:
```
			v, err := def.Validator.Validate(value)
			if err != nil {
			    log.Panic("Cannot validate")
			}
			if found && (!oFound || !reflect.DeepEqual(v, oValue)) {
				changes[field] = v
			}
```
Maybe we can refactor `Prepare` to propagate that `error` instead of panic.

Dragomir-Ivanov:
Fields stored in non-scalar form by `Storer` will differ from incoming JSON
scalar form. Input scalars need to be converted to non-scalar form before
comparison, so decision to be made if the field has changed. Closes #232

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -121,10 +121,17 @@ func (s Schema) Prepare(ctx context.Context, payload map[string]interface{}, ori
 			// Handle prepare on an updated document (original provided).
 			oValue, oFound := (*original)[field]
 			// Apply value to change-set only if the field was not identical same in the original doc.
-			if found && (!oFound || !reflect.DeepEqual(value, oValue)) {
-				changes[field] = value
-			}
-			if !found && oFound && replace {
+			if found {
+				if def.Validator != nil {
+					if validated, err := def.Validator.Validate(value); err != nil {
+						changes[field] = value
+					} else if !oFound || !reflect.DeepEqual(validated, oValue) {
+						changes[field] = validated
+					}
+				} else if !oFound || !reflect.DeepEqual(value, oValue) {
+					changes[field] = value
+				}
+			} else if oFound && replace {
 				// When replace arg is true and a field is not present in the payload but is in the original,
 				// the tombstone value is set on the field in the change map so validator can enforce the
 				// ReadOnly and then the field can be removed from the output document.
--- a/rest/method_item_put_test.go
+++ b/rest/method_item_put_test.go
@@ -582,3 +582,84 @@ func TestPutItemRequiredDefault(t *testing.T) {
 		t.Run(n, tc.Test)
 	}
 }
+
+func TestPutItemReadOnly(t *testing.T) {
+	now := time.Now()
+	yesterday := now.Add(-24 * time.Hour)
+
+	timeOldStr := "2018-01-03T00:00:00+02:00"
+	timeOld, _ := time.Parse(time.RFC3339, timeOldStr)
+
+	sharedInit := func() *requestTestVars {
+		s1 := mem.NewHandler()
+		s1.Insert(context.Background(), []*resource.Item{
+			{ID: "1", ETag: "a", Updated: yesterday, Payload: map[string]interface{}{"id": "1", "foo": "odd"}},
+			{ID: "2", ETag: "b", Updated: yesterday, Payload: map[string]interface{}{"id": "2", "foo": "odd", "zar": "old"}},
+			// Storer will persist `schema.Time{}` as `time.Time` type.
+			{ID: "3", ETag: "c", Updated: yesterday, Payload: map[string]interface{}{"id": "3", "foo": "odd", "tar": timeOld}},
+		})
+
+		idx := resource.NewIndex()
+		idx.Bind("foo", schema.Schema{
+			Fields: schema.Fields{
+				"id":  {Sortable: true, Filterable: true},
+				"foo": {Filterable: true},
+				"zar": {ReadOnly: true},
+				"tar": {ReadOnly: true, Validator: &schema.Time{}},
+			},
+		}, s1, resource.DefaultConf)
+
+		return &requestTestVars{
+			Index:   idx,
+			Storers: map[string]resource.Storer{"foo": s1},
+		}
+	}
+
+	tests := map[string]requestTest{
+		`put:read-only:string:new`: {
+			Init: sharedInit,
+			NewRequest: func() (*http.Request, error) {
+				body := bytes.NewReader([]byte(`{"foo": "baz", "zar":"old"}`))
+				return http.NewRequest("PUT", `/foo/1`, body)
+			},
+			ResponseCode: http.StatusUnprocessableEntity,
+			ResponseBody: `{"code":422,"issues":{"zar":["read-only"]},"message":"Document contains error(s)"}`,
+			ExtraTest:    checkPayload("foo", "1", map[string]interface{}{"id": "1", "foo": "odd"}),
+		},
+		`put:read-only:string:old`: {
+			Init: sharedInit,
+			NewRequest: func() (*http.Request, error) {
+				body := bytes.NewReader([]byte(`{"foo": "baz", "zar":"old"}`))
+				return http.NewRequest("PUT", `/foo/2`, body)
+			},
+			ResponseCode: http.StatusOK,
+			ResponseBody: `{"foo":"baz","id":"2","zar":"old"}`,
+			ExtraTest:    checkPayload("foo", "2", map[string]interface{}{"id": "2", "foo": "baz", "zar": "old"}),
+		},
+		`put:read-only:time:old`: {
+			Init: sharedInit,
+			NewRequest: func() (*http.Request, error) {
+				body := bytes.NewReader([]byte(`{"foo": "odd", "tar":"2018-01-03T00:00:00+02:00"}`))
+				return http.NewRequest("PUT", `/foo/3`, body)
+			},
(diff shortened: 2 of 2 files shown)


Dragomir-Ivanov:
I don't recall any breaking changes, except that updating read-only non-scalar fields with the same value will not bail out. Same as updating scalar fields.

smyrman:
The breaking change is the change of the `schema.Prepare` interface and implementation. We must consider any change to the public API a breaking change, no matter if it's _commonly_ used by applications or not.

E.g. we call Prepare directly several places, where we have code working directly against the `resource` layer without going through `rest`.

Dragomir-Ivanov:
Ah, I see now. Didn't considered someone to use `Prepare` outside rest. I will add it to README.

rest/method_item_put.go
```diff
@@ -51,13 +51,20 @@ func itemPut(ctx context.Context, r *http.Request, route *RouteMatch) (status in
...
+		if len(errs) > 0 {
+			return 422, nil, &Error{422, "Document contains error(s)", errs}
+		}
 		status = 201
 	} else {
 		// PUT used to replace an existing document.
-		changes, base = rsrc.Validator().Prepare(ctx, payload, &original.Payload, true)
+		changes, base, errs = rsrc.Validator().Prepare(ctx, payload, &original.Payload, true)
+		if len(errs) > 0 {
+			return 422, nil, &Error{422, "Document contains error(s)", errs}
```
smyrman: One feedbak I got... Since we do not validate _all_ fields as parts of prepare (e.g. only fields present), I suppose this means we are returning a potentially _partial_ error response, right?

smyrman:
> Ah, I see now. Didn't considered someone to use `Prepare` outside rest. I will add it to README.

It's OK to wait untilll we have discussed/agreed on the interface change though 😉  Let's wait for @rs feedback.

rest/method_item_put.go
Dragomir-Ivanov: Well, yes. The remaining validation will be done by `schema.Validate()`. I guess we can fix this in the future rewrite of this part of the system.

smyrman: Sorry for the delay. I think I may have a proposal that is better.

Maybe we could handle the errors in prepare instead of returning it, by assuming that if an input value fails validation, we treat it as _changed_, and places the original input value in `changes` without a DeepEqual test. We document that in a comment. Then we get to return all the errors to the user, because the same value would fail `schema.Validate()` later anyway.

schema/schema.go
```diff
@@ -99,9 +99,10 @@ func (s Schema) GetField(name string) *Field {
 // being absent). This instruct the validator that the field has been edited, so
 // ReadOnly flag can throw an error and the field will be removed from the
 // output document. The OnInit is also called instead of the OnUpdate.
-func (s Schema) Prepare(ctx context.Context, payload map[string]interface{}, original *map[string]interface{}, replace bool) (changes map[string]interface{}, base map[string]interface{}) {
+func (s Schema) Prepare(ctx context.Context, payload map[string]interface{}, original *map[string]interface{}, replace bool) (changes map[string]interface{}, base map[string]interface{}, errs map[string][]interface{}) {
```
smyrman: Basically revert this ☝️

schema/schema.go
```diff
@@ -99,9 +99,10 @@ func (s Schema) GetField(name string) *Field {
...
 	changes = map[string]interface{}{}
 	base = map[string]interface{}{}
+	errs = map[string][]interface{}{}
```
smyrman: remove

schema/schema.go
```diff
@@ -121,8 +122,17 @@ func (s Schema) Prepare(ctx context.Context, payload map[string]interface{}, ori
...
+				if def.Validator != nil {
+					var err error
+					if value, err = def.Validator.Validate(value); err != nil {
+						addFieldError(errs, field, err.Error())
+						continue
+					}
+				}
+				if !oFound || !reflect.DeepEqual(value, oValue) {
+					changes[field] = value
+				}
```
smyrman: Here; treat errors as a s change 😉 

```go
if val := def.Validator; found && val != nil {
	// We treat a validation error as a change; the validation error
	// indicate invalid payload and will be caught again by Validate().
	if validated, err := val.Validate(value); err != nil || !oFound || !reflect.DeepEqual(validated,  oValue) {
		changes[field] = validated
	}
} else if found {
	if refllect.DeepEqual(value, oValue) {
		changes[field] = value
	}
} else if oFound && replace {
	// When replace arg is true ...
	...
}
```

schema/schema.go
```diff
@@ -158,7 +168,11 @@ func (s Schema) Prepare(ctx context.Context, payload map[string]interface{}, ori
 				if subPayload, ok := value.(map[string]interface{}); ok {
 					// If payload contains a sub-document for this field, validate it
 					// using the sub-validator.
-					c, b := def.Schema.Prepare(ctx, subPayload, subOriginal, replace)
+					c, b, subErrs := def.Schema.Prepare(ctx, subPayload, subOriginal, replace)
+					if len(subErrs) > 0 {
+						addFieldError(errs, field, subErrs)
+						continue
+					}
```
smyrman: can be reverted

schema/schema.go
```diff
@@ -167,7 +181,11 @@ func (s Schema) Prepare(ctx context.Context, payload map[string]interface{}, ori
 			} else {
 				// If the payload doesn't contain a sub-document, perform validation
 				// on an empty one so we don't miss default values.
-				c, b := def.Schema.Prepare(ctx, map[string]interface{}{}, subOriginal, replace)
+				c, b, subErrs := def.Schema.Prepare(ctx, map[string]interface{}{}, subOriginal, replace)
+				if len(subErrs) > 0 {
+					addFieldError(errs, field, subErrs)
+					continue
+				}
```
smyrman: can be reverted

smyrman:
This also means there will be no breaking changes.

Dragomir-Ivanov:
On it.

schema/schema.go
Dragomir-Ivanov: I think this snipped is wrong, stating that if values are the same, then it is a change:
```
} else if found {
	if refllect.DeepEqual(value, oValue) {
		changes[field] = value
	}
```
Perhaps we need to use the same expression as above:
```
} else if !oFound || !reflect.DeepEqual(value,  oValue) {
		changes[field] = value
	}
```

Dragomir-Ivanov:
@smyrman I have resolved all conversations and rebased the PR.

smyrman:
Going ahead with the merge then.

PR #237, opened 2019-02-15
Fix filtering on array containing an object

Dragomir-Ivanov:
This patch allows using `filter={field.subfield:"foo"}` or
`filter={field.subfield:{$regex:"^foo"}}` where `field` is an array
containing an object, and `subfield` is object field.

Also make `rest-layer-mem` match Mongo behaviour in this usecase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -919,20 +919,46 @@ The same example with flags:
 However, keep in mind that Storers have to support regular expression and depending on the implementation of the storage handler the accepted syntax may vary.
 An error of `ErrNotImplemented` will be returned for those storage backends which do not support the `$regex` operator.
 
+The `$elemMatch` operator matches documents that contain an array field with at least one element that matches all the specified query criteria.
+```go
+			"telephones": schema.Field{
+				Filterable: true,
+				Validator: &schema.Array{
+					Values: schema.Field{
+						Validator:  &schema.Object{Schema: &Telephone},
+					},
+				},
+			},
+```
+
+Matching documents that contain specific values within array objects can be done with `$elemMatch`:
+```js
+{telephones: {$elemMatch: {name: "John Snow", active: true}}}
+```
+The snippet above will return all documents, which `telephones` array field contains objects that have `name` **_AND_** `active` fields matching queried values.
+> Note that documents returned may contain other objects in `telephones` that don't match the query above, but at least one object will do. Further filtering could be needed on the API client side.
+
+#### *$elemMatch* Limitation
+`$elemMatch` will work only for arrays of objects for now. Later it could be extended to work on plain arrays e.g:
+```js
+{numbers: {$elemMatch: {$gt: 20}}}
+```
+
 #### Filter operators
 
-| Operator  | Usage                           | Description
-| --------- | ------------------------------- | ------------
-| `$or`     | `{$or: [{a: "b"}, {a: "c"}]}`   | Join two clauses with a logical `OR` conjunction.
-| `$and`    | `{$and: [{a: "b"}, {b: "c"}]}`  | Join two clauses with a logical `AND` conjunction.
-| `$in`     | `{a: {$in: ["b", "c"]}}`        | Match a field against several values.
-| `$nin`    | `{a: {$nin: ["b", "c"]}}`       | Opposite of `$in`.
-| `$lt`     | `{a: {$lt: 10}}`                | Fields value is lower than specified number.
-| `$lte`    | `{a: {$lte: 10}}`               | Fields value is lower than or equal to the specified number.
-| `$gt`     | `{a: {$gt: 10}}`                | Fields value is greater than specified number.
-| `$gte`    | `{a: {$gte: 10}}`               | Fields value is greater than or equal to the specified number.
-| `$exists` | `{a: {$exists: true}}`          | Match if the field is present (or not if set to `false`) in the item, event if `nil`.
-| `$regex`  | `{a: {$regex: "fo[o]{1}"}}`     | Match regular expression on a field's value.
+| Operator     | Usage                           | Description
+| -------------| --------------------------------| ------------
+| `$or`        | `{$or: [{a: "b"}, {a: "c"}]}`   | Join two clauses with a logical `OR` conjunction.
+| `$and`       | `{$and: [{a: "b"}, {b: "c"}]}`  | Join two clauses with a logical `AND` conjunction.
+| `$in`        | `{a: {$in: ["b", "c"]}}`        | Match a field against several values.
+| `$nin`       | `{a: {$nin: ["b", "c"]}}`       | Opposite of `$in`.
+| `$lt`        | `{a: {$lt: 10}}`                | Fields value is lower than specified number.
+| `$lte`       | `{a: {$lte: 10}}`               | Fields value is lower than or equal to the specified number.
+| `$gt`        | `{a: {$gt: 10}}`                | Fields value is greater than specified number.
+| `$gte`       | `{a: {$gte: 10}}`               | Fields value is greater than or equal to the specified number.
+| `$exists`    | `{a: {$exists: true}}`          | Match if the field is present (or not if set to `false`) in the item, event if `nil`.
+| `$regex`     | `{a: {$regex: "fo[o]{1}"}}`     | Match regular expression on a field's value.
+| `$elemMatch` | `{a: {$elemMatch: {b: "foo"}}}` | Match array items against multiple query criteria.
 
 *Some storage handlers may not support all operators. Refer to the storage handler's documentation for more info.*
 
--- a/rest/method_get_test.go
+++ b/rest/method_get_test.go
@@ -343,3 +343,255 @@ func TestGetListFilter(t *testing.T) {
 		t.Run(n, tc.Test)
 	}
 }
+
+func TestGetListArray(t *testing.T) {
+	sharedInit := func() *requestTestVars {
+		s := mem.NewHandler()
+		s.Insert(context.TODO(), []*resource.Item{
+			{ID: "1", Payload: map[string]interface{}{"id": 1,
+				"foo": []interface{}{
+					map[string]interface{}{
+						"a": "bar",
+						"b": 10,
+					},
+					map[string]interface{}{
+						"a": "bar1",
+						"b": 101,
+					},
+				}},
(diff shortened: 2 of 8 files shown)


Dragomir-Ivanov:
@smyrman This is my premature solution for this issue. I still have to add test cases for the fix, however I first wanted to gather some feedback from you.

smyrman:
FYI, I just learned that GitHub has support for opening _draft PRs_ now.

https://github.blog/2019-02-14-introducing-draft-pull-requests/

Probably nice to open one of those in the future for preliminary work ;-)

smyrman:
> This patch allows using `filter={field.subfield:"foo"}` or
`filter={field.subfield:{$regex:"^foo"}}` where `field` is an array
containing an object, and `subfield` is object field.

Should probably follow this syntax?

https://docs.mongodb.com/manual/tutorial/query-array-of-documents/

``filter={field:{subfield:{$regex:"^foo"}}}``, `filter={field:{subfield:"foo"}}`

smyrman:
Actually, the syntax you describe is still supported by MongoDB. I missed that at first. We are implementing a sub-set of the MongoDB syntax through rest-layer, and therefore can pick and choose one alternative when there are several to choose from.

I think perhaps the one you have picked would be the simplest one to use 👍

Dragomir-Ivanov:
`filter={"telephones":{"value":{"$regex":"123"}}}` doesn't work currently. Gives:
```
{
  "code": 422,
  "issues": {
    "filter": [
      "telephones: invalid query expression: not an array"
    ]
  },
  "message": "URL parameters contain error(s)"
}
```
Will try investigate when have some time.

smyrman:
I guess our case was to match several values on the same array entry.

Dragomir-Ivanov:
Hi @smyrman , still no advice from @torie , so I guess it is just two of us. 
I have added a bunch of test cases, as well fixed `rest-layer-mem` `Match()` for `gt, gte, lt, lte`.

smyrman:
I will have a closer look tomorrow.

smyrman:
Here is the thing I wanted feedback on, but I will explain it instead.

Consider we a schema that stores sellers, and their current item stock. Each item has a _color_ and a _price_. Not a great real-world example, but it should be conceptually simple to understand.

Now, let's say the DB has two objects:
```json
{
  "seller": "Tim",
  "items": [
    {"color": "red", "price":"5"},
    {"color": "green", "price":"10"}
  ]
},
{
  "seller": "Jade",
  "items": [
    {"color": "green", "price":"5"},
    {"color": "red", "price":"15"}
  ]
}
```

If we want to find all sellers that have _green_ items that cost _less than 7_, the dotted query syntax won't allow it. This query would match both `Tim` and Jade`, because both have green items, and both have items with a price less than 7:

```js
{items.color:"green", items.price: {$lt: 7}}
```

To construct a query that gives the desired results, we must use one of the following equivalent MongoDB queries:
```js
{items={color:"green", price:{$lt:7}}
{items={$elemMatch: {color:"green", price:{$lt:7}}}
```
This will match onle Jade.

So, therefore, the query you have implemented have some limitation in usage, although it's easier to use when you don't need to query for multiple fields.

We have basically three alternatives:
1. Allow only the dotted query syntax you suggest
2. Allow only the query syntax with `$elemMatch`
3 Allow both

1 is less expressive but simpler to use when it's all you need.
2 allows more expressive queries.
3 makes it harder to write storage backends (more to do/consider).

I am thinking out load, that perhaps the best option is to go for 2 with the explicit `$elemMatch` keyword. The reason is that why MongoDB can handle all of these queries, for other (future) backends, it's probably easiest to handle the explicit keyword.

Does it make sense?

Dragomir-Ivanov:
Thanks @smyrman ,It makes perfect sense. In fact I doubt current `array.field` form is supported by any other driver than Mongo. I can't promise any time-frame, but I am devoted implementing this. Can you add me ass `assignee` for this issue.

Dragomir-Ivanov:
@smyrman Latest commit implements `$elemMatch` in `rest-layer`. If you approve this, I will add needed part in `rest-layer-mongo` and possibly `rest-layer-sql`, and will remove my previous work with `array.field` accessor.

smyrman:
Will have a look tomorrow. Thanks for you work!

schema/query/predicate.go
```diff
@@ -545,3 +547,69 @@ func (e *Regex) Prepare(validator schema.Validator) error {
...
+		if v, ok := val.(map[string]interface{}); ok {
+			match := true
+			for _, subQuery := range e.Exps {
+				if !subQuery.Match(v) {
+					match = false
+				}
+			}
+			if match {
+				return true
+			}
```
smyrman: Does it work to just do this?

```go
p := Predicate: Predicate(e.Exps)
for _, val := range arr {
    if p.Match(val) {
        return true
    }
}
```

schema/query/predicate.go
```diff
@@ -545,3 +547,69 @@ func (e *Regex) Prepare(validator schema.Validator) error {
...
+	if arr, ok = value.([]interface{}); !ok {
+		return false
+	}
+
+	if e.Exps == nil || len(e.Exps) == 0 {
+		// nil or empty predicates will not match
+		return false
+	}
+	for _, val := range arr {
+		if v, ok := val.(map[string]interface{}); ok {
```
smyrman: It doesn't need to be an object for the query to match.

Examples of legal queries for an array of numbers for instance:
```js
{items: {$elemMatch: {$lt: 32}}}
{items: {$elemMatch: {$and:[{$lt: 32},{$gt: 5}]}}
```

In MongoDB, the latter one would be written more tidily using _implicit_ and `{items: {$elemMatch: {$lt: 32, $gt: 5}}}`, but that's does not currently work for comparison queries through rest-layer due to the way queries are parsed.

The docs states that `$elemMatch` is _not_ needed, if matching a single criteria, but it might be good to check weather or not it works. E.g.:

```js
{items: {$elemMatch: 6}}
```
If it works, that makes things easier.

README.md
```diff
@@ -919,20 +919,40 @@ The same example with flags:
 However, keep in mind that Storers have to support regular expression and depending on the implementation of the storage handler the accepted syntax may vary.
 An error of `ErrNotImplemented` will be returned for those storage backends which do not support the `$regex` operator.
 
+Having an `schema.Array` field with values being `schema.Object`:
```
smyrman: Should probably rewrite this a bit.

> The `$elemMatch` operator matches documents that contain an array field with at least one element that matches all the specified query criteria.

No mention that the element needs to be an object.

README.md
```diff
@@ -919,20 +919,40 @@ The same example with flags:
...
+| `$and`       | `{$and: [{a: "b"}, {b: "c"}]}`  | Join two clauses with a logical `AND` conjunction.
+| `$in`        | `{a: {$in: ["b", "c"]}}`        | Match a field against several values.
+| `$nin`       | `{a: {$nin: ["b", "c"]}}`       | Opposite of `$in`.
+| `$lt`        | `{a: {$lt: 10}}`                | Fields value is lower than specified number.
+| `$lte`       | `{a: {$lte: 10}}`               | Fields value is lower than or equal to the specified number.
+| `$gt`        | `{a: {$gt: 10}}`                | Fields value is greater than specified number.
+| `$gte`       | `{a: {$gte: 10}}`               | Fields value is greater than or equal to the specified number.
+| `$exists`    | `{a: {$exists: true}}`          | Match if the field is present (or not if set to `false`) in the item, event if `nil`.
+| `$regex`     | `{a: {$regex: "fo[o]{1}"}}`     | Match regular expression on a field's value.
+| `$elemMatch` | `{a: {$elemMatch: {b: "foo"}}}` | Match array object's field against value.
```
smyrman: Match array items against multiple query criteria.

schema/query/predicate.go
Dragomir-Ivanov: Well `p.Match` needs `map[string]interface{}` and plain `interface{}` will not work. Apart from that good catch +1.

schema/query/predicate.go
Dragomir-Ivanov: Yes, I guess `$elemMatch` can be more versatile. I will think a bit more, on how to make it work on array of values, instead of array of objects.

README.md
Dragomir-Ivanov: Done.

README.md
Dragomir-Ivanov: Done.

schema/query/predicate.go
Dragomir-Ivanov: The reason for mandatory `schema.Object`:
```
	if obj, ok = arr.Values.Validator.(*schema.Object); !ok {
		return fmt.Errorf("%s: array elements are not schema.Object", e.Field)
	}
	return prepareExpressions(e.Exps, obj.Schema)
```
Is that `prepareExpressions` needs `schema.Validator`, and not `schema.FieldValidator`.

smyrman: Ok, thanks for the explanation.

Let's keep it for objects only for now, and document it as a _current_ limitation.

Dragomir-Ivanov:
Hi @smyrman , this PR is rebased and ready for merge.

Dragomir-Ivanov:
Just added second commit that fixes an issue with filtering on `schema.Array`. I guess it can pass in the same PR.

rest/method_get_test.go
```diff
@@ -343,3 +343,314 @@ func TestGetListFilter(t *testing.T) {
 		t.Run(n, tc.Test)
 	}
 }
+
+func TestGetListArray(t *testing.T) {
+	sharedInit := func() *requestTestVars {
+		s := mem.NewHandler()
+		s.Insert(context.TODO(), []*resource.Item{
```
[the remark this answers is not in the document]
Dragomir-Ivanov: Thanks, I was just following what was being used before, mostly copy-paste from previous tests.

Dragomir-Ivanov:
Thank you for your time @smyrman . I really appreciate your advises. 
Done with requested changes.

Dragomir-Ivanov:
@smyrman Added another small commit fixing #240.

PS: That book of Dave Chaney is awesome!

smyrman:
@Dragomir-Ivanov, will merge this now. Thanks for you work 🙌

PR #241, opened 2019-03-08
Fix field hook for HTTP PUT

Dragomir-Ivanov:
Always calls `OnUpdate` field hook on HTTP PUT for existing documents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ Below is an overview over recent breaking changes, starting from an arbitrary po
   - `filter` parameters will be validated for type match only, instead of type & constrains.
 - PR #228: `Reference` projection fields will be validated against referenced resource schema.
 - PR #230: `Connection` projection fields will be validated against connected resource schema.
+- PR #241: Always call `OnUpdate` field hook on HTTP PUT for existing documents.
 
 From the next release and onwards (0.2), this list will summarize breaking changes done to master since the last release.
 
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Below is an overview over recent breaking changes, starting from an arbitrary po
   - `filter` parameters will be validated for type match only, instead of type & constrains.
 - PR #228: `Reference` projection fields will be validated against referenced resource schema.
 - PR #230: `Connection` projection fields will be validated against connected resource schema.
-- PR #241: Always call `OnUpdate` field hook on HTTP PUT for existing documents.
+- PR #241: Always call `OnUpdate` field hook on HTTP PUT for existing documents. Deleting a field with `Default` value set, will always be reset to its default value.
 
 From the next release and onwards (0.2), this list will summarize breaking changes done to master since the last release.
 
@@ -1252,7 +1252,9 @@ Used to query a resource with its sub/embedded resources.
 Used to create new resource document, where new `ID` is generated from the server.
 
 ### PUT
-Used to create/update single resource document given its `ID`. Be aware when dealing with resource fields with `Default` set. Initial creation for such resources will set particular field to its default value if omitted, however on subsequent `PUT` calls this field will be deleted if omitted. If persistent `Default` field is needed use `{Required: true}` with it. 
+Used to create/update single resource document given its `ID`. On initial request, `onInsert` resource hook will be called. All subsequent `PUT` requests will call `onUpdate` resource hook.
+
+Be aware when dealing with resource fields with `Default` set. If missing from payload or deleted, such fields will be reset to their `default value`.
 
 ### PATCH
 Used to update/patch single resource document given its `ID`. REST Layer supports following update protocols:
--- a/rest/method_item_put_test.go
+++ b/rest/method_item_put_test.go
@@ -388,8 +388,8 @@ func TestPutItemDefault(t *testing.T) {
 				return http.NewRequest("PUT", "/foo/2", body)
 			},
 			ResponseCode: http.StatusOK,
-			ResponseBody: `{"id": "2", "foo": "baz"}`,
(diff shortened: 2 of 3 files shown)


smyrman:
@rs, I still think that calling `OnUpdate` is the most _useful_ thing for the `PUT` request (or the `resource` level call) to do in _most cases_, and it's not any less _idempotent_ than calling `OnIsert`. As I stated before, a REST-API with hooks, can never grantee _idempotent_ behaviour.

Common cases where I think calling `OnUpdate` is better, is fields such as `updatedAt` and `createdAt`.

Where I see it would be _correct_ to call an `OnInit` hook, would be if the _hook_ was responsible of setting the field default value. Since we have the dedicated `Default` fields, the cases where this is
needed is limited. @Dragomir-Ivanov, we should still always reset non-present fields to the Default value on _replace_ though (disregarding the state of the `required` field, as I think different behavior there just causes confusion).

Perhaps we can side-step this disagreement by adding an `OnReplace` field hook that takes _precedence_ over `OnUpdate` when explicitly set only? That way the API user is in full control over which hook to use for PUT and JSON-Patch requests, while the default behavior remain useful.

I want to state, that for the `resource` level hooks, it's probably the most idempotent option to call `OnUpdate` / `OnUpdated` hooks for replace operations when a previous resource exists.

## Example use-case of idemponent behavior (PUT)

Let's look at an example use-case that I like to come back to, where _idemponent_ behavior is important:

A _telemetry_ system running remotely, is reporting some data for a server. Let's say it's connected to a smart-home system. Before it reports the data, it _ensures_ a resource for each device in rest-layer; the resource is populated with fields from a _config file_. The first time a new device is added, a user-editable _item_ is added to expose the device in the UI. The _item_ is used to override names, units and other values.

Now, for each restart of the remote telemetry software, I would want the _device_ entry fields that are removed from the config file to be reset to their defaults or cleared and new or altered fields to be updated. I also want the device `updatedAt` time to be updated, but I _don't_ want the device's `createdAt` time to be altered. I _certainly_ don't want another _item_ to be created, as that would just be confusing.

In my minds, that means that the best  _idemponent_ behavior in this case is achieved with:
- Reset to `default` values for omitted fields with defaults.
- Call `onUpdate` hooks on `Update` calls with `replace` set to true.
- Call `Update`, `Updated` resource hooks.

rs:
Sounds good as long as we consistently call the same hook whatever the previous state of the resource is for PUT.

Dragomir-Ivanov:
Hi @smyrman, made another patch, which will always put `Default` value on deleted field.

smyrman:
> Sounds good as long as we consistently call the same hook whatever the previous state of the resource is for PUT.

The idea would be _not_ to do that. `createdAt` times should be set at insert only, and `updatedAt` times should be set at update & replace time only. Anything else would be _very_ confusing, and potentially unsafe. I would go as far as to say that the _integrity_ of the data has been lost; these are normally _read-only_ fields, and it's clearly not _intended_ that it should be allowed to modify them in this way. Perhaps deletion of items in this resources is _not allowed_, and the read-only `createTime` is expected to be trustworthy. The fields could contain other values that _must_ be set correctly on insert and should not be altered on update, such as _permissions_. 

If you call the same hook every time on `PUT`, that either menas that it's either possible to _create_ an object via `PUT` that you cannot do via `POST` (e.g. `OnInit` not. called), or that you can _update_ an object via `PUT` in a way that you cannot update it via `PATCH` (e.g. `OnUpdate` not called, or `OnInit` called when it shouldn't be). This way it becomes possible to use `PUT` to alter objects in a way that they were never meant to allow. Note that some of this fields could be read-only.

The most important thing to me for an _idempotent_ call in practice, is that it's _safe_ to call several times; that's not the case when it calls the same hook on creation and update.

We have been forced to _disable_ `PUT` in our own APIs for exactly that reason; it's not safe to use!

For a _new_ hook, `OnReplace`, that could take _precedence_ over `OnUpdate` on relace, and it could be _ignored_ for non-replace  updates (`PATCH` requests), e.g. when there is no `OnUpdate` hook. But if you want the same hook to be called on initial creation, I think the programmer would need to set `OnInit` and `OnReplace` to the same value. This way the _intent_ is explicit.

That said, we at least, don't have a use-case for `OnReplace`; `Default` is more than enough for us atm.

smyrman:
> For a new hook, OnReplace, that could take precedence over OnUpdate on replace and it could be ignored for non-replace updates (PATCH requests) 

Better still perhaps, If you want an `OnUpdate` to run `OnReplave`, you need to set `OnReplace` to the same value as `OnUpdate`; that way the _resource_ layer hooks API becomes very straight forward and "stupid" (which is a good thing).

Dragomir-Ivanov:
I am with @smyrman on this one. They way as I understand REST (and my peer friends) is that once you have a resource in the system, with its ID field, all other modifications to that resource are `Updates`. 
If one wants new resource, he can delete current one, and `POST` a new one, getting new ID in the process. The REST PUT term `replaces` is what the system does, but logically from user perspective it is update.

rs:
I see your point. If you both agrees with that, I'm good with it.

smyrman:
Thanks @rs.

With that I think we should go ahead with this @Dragomir-Ivanov. Might be good to update `PUT` documentation to make it perfectly clear how it behaves in terms of hooks and defaults.

Having used rest-layer for a while now, I worry about API design, but I worry more about the data that gets stored, and the states it can be in😏

Dragomir-Ivanov:
Thanks @rs, @smyrman .
Made requested changes to README and rebased.

smyrman:
> This branch is out of sync with the base branch.

Did you do a git fetch before the rebase?

Dragomir-Ivanov:
@smyrman Right, sorry about that. Upstream fetched & rebased.

Issue #243, opened 2019-04-21
schema.VerifyPassword not reliable


markostojanovic087:
I tried implementing basic authentication by using example given in [_basic-authentication_](https://github.com/rs/rest-layer/blob/master/examples/auth/main.go).
As soon as I changed test mem handlers to sql handlers

`sqlHandler, err := sqlStorage.NewHandler(DB_DRIVER, DB_SOURCE, DB_TABLE_USER)`

I started getting "Please provide proper credentials" response.

After long investigation and debug
I figured out that the method schema.VerifyPassword seems to be problematic one.
As soon as I used the following method instead of VerifyPassword I had no issue.

```
func MyVerifyPassword(hash interface{}, password []byte) bool { 
    h, ok := hash.(string)
    if !ok {
        return false
    }
    return bcrypt.CompareHashAndPassword([]byte(h), password) == nil
}
```

The calling code remained exactly the same,
just instead of schema.VerifyPassword I called MyVerifyPassword.

```
// NewBasicAuthHandler handles basic HTTP auth against the provided user resource
func NewBasicAuthHandler(users *resource.Resource) func(next http.Handler) http.Handler {
	return func(next http.Handler) http.Handler {
		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			if u, p, ok := r.BasicAuth(); ok {
				// Lookup the user by its id
				ctx := r.Context()
				user, err := users.Get(ctx, u)

				if user != nil && err == resource.ErrForbidden {
					// Ignore unauthorized errors set by ourselves
					err = nil
				}

				if err != nil {
					// If user resource storage handler returned an error, respond with an error
					if err == resource.ErrNotFound {
						http.Error(w, "Invalid credential", http.StatusForbidden)
					} else {
						http.Error(w, err.Error(), http.StatusInternalServerError)
					}
					return
				}

				if MyVerifyPassword(user.Payload["password"], []byte(p)) {
					// Store the auth user into the context for later use
					r = r.WithContext(NewContextWithUser(ctx, user))
					next.ServeHTTP(w, r)
					return
				}
			}
			// Stop the middleware chain and return a 401 HTTP error
			w.Header().Set("WWW-Authenticate", `Basic realm="API"`)
			http.Error(w, "Please provide proper credentials", http.StatusUnauthorized)
		})
	}
}
```

smyrman:
@markostojanovic087, are you open for doing a PR on this?

markostojanovic087:
[...]

PR #248, opened 2019-04-30
Fix schema.VerifyPassword type assertion

markostojanovic087:
Fixing issue [Issue-243](https://github.com/rs/rest-layer/issues/243)

### DIFF
--- a/schema/password.go
+++ b/schema/password.go
@@ -58,9 +58,9 @@ func (v Password) Validate(value interface{}) (interface{}, error) {
 // VerifyPassword compare a field of an item payload containing a hashed
 // password with a clear text password and return true if they match.
 func VerifyPassword(hash interface{}, password []byte) bool {
-	h, ok := hash.([]byte)
+	h, ok := hash.(string)
 	if !ok {
 		return false
 	}
-	return bcrypt.CompareHashAndPassword(h, password) == nil
+	return bcrypt.CompareHashAndPassword([]byte(h), password) == nil
 }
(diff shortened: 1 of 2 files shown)


markostojanovic087:
I think now that examples provided will fail since they have this code:
```
{ID: "admin", Updated: time.Now(), ETag: "abcd", Payload: map[string]interface{}{
			"id":       "admin",
			"name":     "Dilbert",
			"password": secret,
		}},
```
As you can see the password isn't quoted. I think we should change it also.
@smyrman Do you agreee?

By the way, there is another mistake in this examples which makes some of it's code useless.

Please see the [line](https://github.com/rs/rest-layer/blob/3f2cd5df9e8e6690c56f806710ef53048109c323/examples/auth/main.go#L95).
The correct line will be 
```
if err != nil {
```

The way it is, the code always returns.
@smyrman Should we open another issue or make this change along with this one?

schema/password.go
```diff
@@ -58,9 +58,9 @@ func (v Password) Validate(value interface{}) (interface{}, error) {
 // VerifyPassword compare a field of an item payload containing a hashed
 // password with a clear text password and return true if they match.
 func VerifyPassword(hash interface{}, password []byte) bool {
-	h, ok := hash.([]byte)
+	h, ok := hash.(string)
 	if !ok {
 		return false
 	}
-	return bcrypt.CompareHashAndPassword(h, password) == nil
```
smyrman: I want to suggest this instead:
```go
var h []byte
switch hashT := hash.(type) {
case []byte:
    b = hashT
case string:
    b = []byte(hashT)
default:
    return false
}
return bcrypt.CompareHashAndPassword(h, password) == nil
```

Then we handle both a `[]byte` and a `string` DB value, which both probably can be the case, depending on schema and/or backend.

smyrman:
> @smyrman Should we open another issue or make this change along with this one?

Maybe we can do a separate PR to update any issues in the example / README, then this can be merged sooner.

smyrman:
Closing due to lack of activity.

Issue #255, opened 2019-08-19
OnInsert/OnUpdate hooks can modify the new resource item, Etag not recalculated


Dragomir-Ivanov:
During POST/PUT/PATCH new resource item Etag generation is done in `item, err := resource.NewItem(doc)` from the document supplied from the request. However after that hooks will be applied, that may or may not modify this item further. Modifying the item in the hook, doesn't recalculate the Etag. 
One obvious solution suggested by @smyrman , is moving Etag calculation right before hitting the Store.

smyrman:
Let's try to find a few problem cases, to see how they are handled by any given solution:

E.g. what happens if:
- A resource with an `updateAt` timestamp field is PATCHED with no changes. Is the resource (and hence the E-tag) updated?

Dragomir-Ivanov:
Okay, if you are referring to the invisible `_updated` field, that is produced by `rest-layer` then no, it is not included in the Etag sum. 
```
func NewItem(payload map[string]interface{}) (*Item, error) {
	id, found := payload["id"]
	if !found {
		return nil, errors.New("Missing ID field")
	}
	etag, err := genEtag(payload)
	if err != nil {
		return nil, err
	}
	item := &Item{
		ID:      id,
		ETag:    etag,
		Updated: time.Now(),
		Payload: payload,
	}
	return item, nil
}
```
However if you are including your own:
`"updated": schema.UpdatedField,`, I am afraid it is modification.

smyrman:
I am referring to when you include your own field; or indeed any other field that is _generated_ (by a field or resource hook) on update.

A possible solution for that is to abort an update if there are no E-tag changes after e.g. a patch is applied. This of course need to be _documented_. The next question is -- is there a case where this isn't what we would want?

Another important question is:
- Is it _important_ for a hook to get an updated/accurate E-tag?

smyrman:
Example trade-off:

> However if you are including your own: `"updated": schema.UpdatedField,`, I am afraid it is a modification.

We can avoid this by aborting the update if there are no changes to a resource after a patch or put operation.

Trade-off: It won't be possible to _force_ resource updates (via hooks) by updating (without changing) a field value.

What ever we end up on, the solution must of course be documented. Writing the documentation will also be a good test for the solution; if we can't easily formulate the behaviour in words, probably our solution is too complex.

Dragomir-Ivanov:
[...]

PR #257, opened 2019-08-19
Recalculate ETag after OnUpdate/OnInsert hooks, before storing. Fixes…

Dragomir-Ivanov:
… #255

### DIFF
--- a/resource/resource.go
+++ b/resource/resource.go
@@ -354,10 +354,30 @@ func (r *Resource) Insert(ctx context.Context, items []*Item) (err error) {
 		}(time.Now())
 	}
 	if err = r.hooks.onInsert(ctx, items); err == nil {
-		err = r.storage.Insert(ctx, items)
+		if err = recalcEtag(items); err == nil {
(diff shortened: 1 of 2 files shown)


Dragomir-Ivanov:
About documentation...since this behavior sound common sense, do I still need to add some clarification in the README?

smyrman:
Probably it's OK without additional documentation at this stage.

smyrman:
@Dragomir-Ivanov, do you consider this ready for merge?

smyrman:
If you can, maybe improve the commit message according to these conventions:
- https://chris.beams.io/posts/git-commit/

Especially the subject line. E.g.

```
Recalculate E-Tag right before before storing items

E-tags used to be set after parsing an item only, before potential hooks
alter the item content. This change adds a final step to resource Update
and Create operations to recalculate E-tags right before passing items
to the Storer back-end.

Fixes #257 
```

Or something like that

Dragomir-Ivanov:
Thanks for the suggestion @smyrman. Recommitted with the new message & ready for merge.

PR #263, opened 2019-11-26
Return actual empty response. Fixes #262

### LINKED ISSUE #262: Bogus error line, when returning empty response on 204
When `Prefer: return=minimal` is used, `rest-layer` will return empty body with HTTP code of 204,
however this results in a bogus log line with following text:
```
Can't send response: http: request method or response status code does not allow body
```
Mental trace of the issue:
`func itemPatch` executes this statements:
```
status = 200
	if isNoContent(r) {
		item.Payload = nil
		status = 204
	}
```
thus making `item.Payload = nil`
However later down the pipeline in `func FormatItem(...)(ctx context, body interface{})` we have:
```
	if skipBody {
		return ctx, nil
	}
	return ctx, i.Payload
}
```
If `skipBody` is not flagged, we are returning `map[string]interface{}{nil}` which is not the same as `interface{}{nil}`.
Then later we have `func (s DefaultResponseSender) Send(...)`
```
	w.WriteHeader(status)

	if body != nil {
		j, err := json.Marshal(body)
		if err != nil {
			w.WriteHeader(500)
			logErrorf(ctx, "Can't build response: %v", err)
			msg := fmt.Sprintf("Can't build response: %q", err.Error())
			w.Write([]byte(fmt.Sprintf("{\"code\": 500, \"msg\": \"%s\"}", msg)))
			return
		}
		if _, err = w.Write(j); err != nil {
			logErrorf(ctx, "Can't send response: %v", err)
		}
	}
```
Which checks the `body iterface{}`, and it is not `nil` but `map[string]interface{}{nil}`, so it proceeds with marshaling it to `null` string. Then on the `w.Write()`.
But since Go's http package doesn't like we just wrote `w.WriteHeader(status)` of `204`, it fails on that `w.Write` call with the error string `Can't send response: http: request method or response status code does not allow body`.

Will make a PR shortly.

Dragomir-Ivanov:
Return actual empty response. Fixes #262

### DIFF
--- a/rest/response.go
+++ b/rest/response.go
@@ -77,7 +77,7 @@ func (f DefaultResponseFormatter) FormatItem(ctx context.Context, headers http.H
 	if !i.Updated.IsZero() {
 		headers.Set("Last-Modified", i.Updated.In(time.UTC).Format("Mon, 02 Jan 2006 15:04:05 GMT"))
 	}
-	if skipBody {
+	if skipBody || i.Payload == nil {
 		return ctx, nil
 	}
 	return ctx, i.Payload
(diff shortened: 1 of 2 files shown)


Dragomir-Ivanov:
Sorry @rs , forgot the tests. Can you approve again? Thanks!

smyrman:
I am going ahed to merge this @rs and @Dragomir-Ivanov.

Issue #273, opened 2021-09-23
Support for $not filter operation


tsetsoo:
Hello,

I am using Mongo DB as storage, but not [rest-layer-mongo](https://github.com/rs/rest-layer-mongo).

I want to implement _Contains_ and _Does Not Contain_ filters. For _Contains_ I can use the `$regex` operation. However, for _Does Not Contain_ I cannot use `$regex`, since Go does not support lookarounds and regexes are currently validated with Go's standard library. Using `$not` seems to me like the obvious alternative.

Per #272 I know that there is not active development on this library, so this issue is to ask if a reason exist for this feature to be missing. If there is no such reason, I will implement it myself (including support for `$not`-ing a regex and `$not`-ing a whole expression)

smyrman:
I don't see any reason why this feature could not be added. Also, please do feel free to create a PR for it.

> I am using Mongo DB as storage, but not rest-layer-mongo.

Want to share what you are using? Is it in a state in which other could use it, or it's written more specifically for the purpose at hand.

tsetsoo:
I've opened #274 with the initial implementation of the `$not` operator with regex, functioning as an opposite to `$regex`.
I'll add a second PR with the implementation of `$not`-ing a whole expression since that seems a bit [trickier](https://github.com/tsetsoo/rest-layer/blob/full-featured/schema/query/predicate.go#L585)

> Want to share what you are using? Is it in a state in which other could use it, or it's written more specifically for the purpose at hand?

Unfortunately, it's written for the purpose at hand. One of the main blocker would have been the support for mgo over [mongo](https://github.com/mongodb/mongo-go-driver).
Regardless, I'll open a PR in rest-layer-mongo after I'm finished here, to add support for `$not`

smyrman:
> $not-ing a whole expression since that seems a b

To be honest, I am not sure how _useful_ it is to be able to not the whole expression. The docs does state that
`{price: {$not: {$gt: 12}}}` is not equivalent to`{price: {$lte: 12}}`, since the prior would match any documents where the price is not set while the latter would not. If this is _really_ the query you want, you could just do `{$or: [{price: {$lte: 12}}, {price:{$exists: false}]}`, which would be more explicit anyways.

smyrman:
So at least I am not against _only_ implement `$not` as an oposite to `$regex`. An yviews from you @rs?

For reference: https://docs.mongodb.com/manual/reference/operator/query/not/

PR #274, opened 2021-09-23
implement '$not' regex

tsetsoo:
Implementing `$not` filter operator with regex. It has the opposite function of `$regex`.

Part of #273

### DIFF
--- a/README.md
+++ b/README.md
@@ -998,6 +998,8 @@ The same example with flags:
 However, keep in mind that Storers have to support regular expression and depending on the implementation of the storage handler the accepted syntax may vary.
 An error of `ErrNotImplemented` will be returned for those storage back-ends which do not support the `$regex` operator.
 
+The operator `$not` functions as an opposite operator to `$regex`. Unlike MongoDB, we do not allow `$not` as a general negation operator.
+
 The `$elemMatch` operator matches documents that contain an array field with at least one element that matches all the specified query criteria.
 ```go
 			"telephones": schema.Field{
@@ -1041,6 +1043,7 @@ The snippet above will return all documents, which `telephones` array field cont
 | `$gte`       | `{a: {$gte: 10}}`               | Fields value is greater than or equal to the specified number.
 | `$exists`    | `{a: {$exists: true}}`          | Match if the field is present (or not if set to `false`) in the item, event if `nil`.
 | `$regex`     | `{a: {$regex: "fo[o]{1}"}}`     | Match regular expression on a field's value.
+| `$not`       | `{a: {$not: "fo[o]{1}"}}`       | Opposite of `$regex`.
 | `$elemMatch` | `{a: {$elemMatch: {b: "foo"}}}` | Match array items against multiple query criteria.
 
 *Some storage handlers may not support all operators. Refer to the storage handler's documentation for more info.*
--- a/schema/query/predicate_parser.go
+++ b/schema/query/predicate_parser.go
@@ -107,7 +107,7 @@ func (p *predicateParser) parseExpression() (Expression, error) {
 		or := Or(subExp)
 		return &or, nil
 	case opExists, opIn, opNotIn, opNotEqual, opRegex, opElemMatch,
-		opLowerThan, opLowerOrEqual, opGreaterThan, opGreaterOrEqual:
+		opLowerThan, opLowerOrEqual, opGreaterThan, opGreaterOrEqual, opNot:
 		p.pos = oldPos
 		return nil, fmt.Errorf("%s: invalid placement", label)
 	default:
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -998,6 +998,8 @@ The same example with flags:
 However, keep in mind that Storers have to support regular expression and depending on the implementation of the storage handler the accepted syntax may vary.
 An error of `ErrNotImplemented` will be returned for those storage back-ends which do not support the `$regex` operator.
 
+The operator `$not` functions as an opposite operator to `$regex`
```
smyrman, line 1001: We should stick to the MongoDB definition if implementing this:

https://docs.mongodb.com/manual/reference/operator/query/not/

README.md
```diff
@@ -1041,6 +1043,7 @@ The snippet above will return all documents, which `telephones` array field cont
 | `$gte`       | `{a: {$gte: 10}}`               | Fields value is greater than or equal to the specified number.
 | `$exists`    | `{a: {$exists: true}}`          | Match if the field is present (or not if set to `false`) in the item, event if `nil`.
 | `$regex`     | `{a: {$regex: "fo[o]{1}"}}`     | Match regular expression on a field's value.
+| `$not`       | `{a: {$not: "fo[o]{1}"}}`       | Opposite of `$regex`.
```
smyrman, line 1046: To get the exact functionality you are after, you would need to do:
`{a: {$not: {$regex: "fo[o]{1}"}}}`

smyrman (changes requested):

README.md
smyrman: Ah.. I now see that `$not: string` is allowed as a short-hand for `$not: {$regex: string}` . I suppose that's fair enough.

README.md
smyrman: Did not see your comment in https://github.com/rs/rest-layer/issues/273#issuecomment-925864211 earlier. Maybe extend the text somewhat:

```suggestion
The operator `$not` functions as an opposite operator to `$regex`. Unlike MongoDB, we do not allow `$not` as a general negation operator.
```

schema/query/predicate_parser.go
```diff
@@ -258,6 +254,16 @@ func (p *predicateParser) parseCommand(field string) (Expression, error) {
...
+		case opNot:
+			re, err := p.parseRegex(label)
+			if err != nil {
+				return nil, err
+			}
+			p.eatWhitespaces()
+			if !p.expect('}') {
+				return nil, fmt.Errorf("%s: expected '}' got %q", label, p.peek())
+			}
+			return &Not{Field: field, Value: re}, nil
```
smyrman, line 266: Maybe we could instead return `$Regex{Field: field, Value: re, Negated: true}`.

This would give a clear path-way for how to support the `$not` operator in general later, if there is a need for that.

tsetsoo: great suggestion, thanks

README.md
tsetsoo: `$not: {$regex: string}` was added in Mongo 4.0.6 so I suppose `$not: string` is the conventional way. Regardless, if we decide to add support for `$not` as a general negation operator, both will be supported.

README.md
tsetsoo: done

tsetsoo:
@smyrman Do you think we can go ahead and merge this PR? The regex functionality is needed regardless of whether we decide to support the ability to _not_ the whole expression.

smyrman:
Let's go ahead. Could you please squash your changes into one commit?

Example:
```
git fetch
git rebase -i origin/master
```
mark second commit as "fixup".

tsetsoo:
Done

smyrman (dismissed):

schema/query/predicate_parser.go
```diff
@@ -247,7 +247,11 @@ func (p *predicateParser) parseCommand(field string) (Expression, error) {
 			if !p.expect('}') {
 				return nil, fmt.Errorf("%s: expected '}' got %q", label, p.peek())
 			}
-			return &Regex{Field: field, Value: re}, nil
+			negated := false
+			if label == opNot {
+				negated = true
+			}
```
smyrman, line 253: can be written without a branch condition
```suggestion
			negated := label == opNot 
```

smyrman: maybe fix this as well (squash it in, then I can merge).

tsetsoo: done

smyrman (approved):

smyrman:
@rs: Seams travis is no longer working, so I can't merge this:

image

rs:
I removed the requirement for travis check before merging waiting for a github actions migration.

PR #275, opened 2021-10-01
Add GitHub Actions

smyrman:
Back-ported from https://github.com/searis/rest-layer
 
 Badges in README not updated.


smyrman:
If you want to add GitHub actions to this repo, here is a starting-point. If you enable GitHub Actions actions for the repo, it will, run, but at least the golang-ci linter task is going to fail.

smyrman:
As mentioned on Slack, we at [Clarify](https://clarify.io) will be maintaining our own fork of rest-layer going forward; it just makes the development faster and easier for us to control. I have got all the tests working there, and fixed all the linter issues, but we haven't yet transitioned to using it.

We are very grateful for rest-layer, and will continue to maintain the fork with bugfixes and minor features.

rs:
Why not moving the main project into an org?

smyrman:
> Why not moving the main project into an org?

It does make sense to discuss this properly. I started a discussion here https://github.com/searis/rest-layer/discussions/2. It's our "selfish" view of what we need from rest-layer going forward, and what we hope to achieve with a fork. Contributions and views are very much welcome.

